PR #28, opened 2021-11-06
Chrome Beta icon.

arun54321:
image

Made it from existing chrome stable icon.


arun54321:
Another version with slightly reduced text size.
image



https://pastebin.com/RMHwH229

PR #81, opened 2021-11-13
Add Kiwi Browser

naveenjohnsonv:


patrykmichalik (changes requested):
I would suggest using the flat logo found on the [Kiwi Browser website](https://kiwibrowser.com) to make the icon more recognizable.

PR #140, opened 2021-11-25
Patch 1

lilyoo226:


patrykmichalik (changes requested):
Sorry for keeping you waiting! Because SmartCaller isn’t a system app, I’d suggest giving it a custom icon that matches the official logo rather than linking it to `@drawable/dialer`.

PR #145, opened 2021-11-26
Added Gb Whatsapp

### LINKED ISSUE #144: Add gbwhatsapp app
com.gbwhatsapp

lilyoo226:
Fixes #144


patrykmichalik:
Thanks for your interest in contributing! Unfortunately, because GBWhatsApp doesn’t appear to be available from reputable sources (such as the Play Store or APKMirror) or be open-source, we won’t be able to add an icon for it. This it to avoid including icons for potentially unsafe or otherwise dubious apps.

PR #150, opened 2021-11-27
Link qcamera3 package name to camera

kalkioss:
BSG GCam added this packagename recently...


wyrgb:
Or okay, it is not feasible to add all the package names, there are a lot of devs, with their unique packagename. So I guess this should be done based on App Label. Right? @patrykmichalik

patrykmichalik:
Icons are linked to apps based on their package names. Using app names instead of package names would be problematic because, for example, app names may be localized into different languages. This pull request looks good to me, so if you’d like to see the camera icon linked to the “com.codeaurora.qcamera3” package, please feel free to reopen the pull request.

kalkioss:
Okay, so let me do a new pr linking other camera packagenames as well

PR #156, opened 2021-11-28
added GCam shamim, AIMP, Sony Music

MohammedAlamkhan:
GCam shamim is widely used gcam mod with gcam icon


arun54321:
Why are you adding apple music icon to other music player apps?

MohammedAlamkhan:
.

PR #187, opened 2021-12-04
Link org.codeaurora.qcamera3 to camera

kalkioss:


kalkioss:
@patrykmichalik Well I saw that packagename "com.android.mgc" linked here is not working with package "com.android.MGC" which is what being used in the GCam application...
So, wouldn't it be better to make icon pack application not case sensitive?

patrykmichalik (approved):

patrykmichalik:
Thanks for contributing, @kalkioss! And sorry for keeping you waiting. Regarding your comment, package names don’t appear to be case-insensitive. It is possible for two apps whose package names differ only in casing to be installed at the same time.

PR #249, opened 2022-01-17
Add icons for Corona-Warn and CovPass

yfunk:
This PR adds icons for the German Covid contact tracing and digital vaccination certificate apps.

## Apps
- [Corona-Warn](https://play.google.com/store/apps/details?id=de.rki.coronawarnapp) (`de.rki.coronawarnapp`)
- [CovPass](https://play.google.com/store/apps/details?id=de.rki.covpass.app) (`de.rki.covpass.app`)

## Additonal Info
The icons are made from SVGs found in the official media materials and aligned inside the "Content Area for most shapes" square.

cov_orientation


yfunk:
The CovPass icon could also work for the [CovPassCheck](https://play.google.com/store/apps/details?id=de.rki.covpass.checkapp) app, which uses the same shape with an inverted color palette in the original. 

Not sure if it would look weird here if the icon was exactly the same, lmk what you think.

PR #279, opened 2022-01-23
Sorted xml alphabetically

JordyEGNL:
The XML is supposed to be sorted alphabetically, but it went wrong somewhere. Fixed that :)


patrykmichalik:
Thanks, @JordyEGNL!

PR #294, opened 2022-01-27
Kalp dev patch more icons

kalp-dev:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Would you be able to add the SVG files for these icons to the `svgs` directory? Cheers!

PR #301, opened 2022-01-29
improve `Retro Music` icon

SVR666:
### Just updated the icon

**Comparison between old and new**

retro


ZhmieXD:
Oh wow you can actually do multiple shades here? I thought only binary colours 😅

SVR666:
> Oh wow you can actually do multiple shades here? I thought only binary colours 😅

Idk, i didn't see any guidelines

PR #329, opened 2022-02-04
 Add Wikipedia, Steam and KeePassDX icons

IAmRuMaks:
### Before:
Before
### After:
After


IAmRuMaks:
The new wikipedia icon:
Wikipedia icon

J-Jamet:
Pretty cool, nice work. I'll reuse your design in the official logo of KeePassDX if that's ok with you. I'll just add the colors.

PR #386, opened 2022-02-13
Link kiwi browser dev to the existing kiwi browser icon

Indranil012:


patrykmichalik:
Thanks, @Indranil012!

PR #389, opened 2022-02-15
mBank CZ, SK and FOSS Browser icons

Golbinex:
Use mBank PL icon for CZ and SK versions as well.

FOSS Browser's icon is very similar existing Jelly one, as you can see in pictures below:
image
image


patrykmichalik (approved):

patrykmichalik:
Thanks, @Golbinex!

PR #397, opened 2022-02-20
add new icon

Autines:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Please keep in mind that for each new icon, two files should be added: an XML file and an SVG file. This is described in [`CONTRIBUTING.md`](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md). It would appear that this PR doesn’t include SVG files for the newly added icons.

PR #420, opened 2022-03-06
Linked Material Files app

Giann1s:


lazerl0rd:
If the review was waiting on a check if the app matches the said icon, the app shares the same icon as AOSP Files (albeit with a slightly darker blue hue as the background) so it should be fine.

PR #436, opened 2022-04-02
Add 19 Icons

fischer-felix:
Added Icons for:
- AMdroid
- CellMapper
- ~~DB Navigator~~
- Flightradar24
- Geizhals
- Microsoft Onenote
- Mobiles Bezahlen
- MVGO
- Notenapp
- OctoApp
- S-pushTAN
- Schulmanager
- Slide for Reddit
- Solar.web
- Sparkasse
- VIMpay
- Wayback Machine
- Wear
- Wetter Online
- Wikipedia

Modified Deezer icon to better fit the design standards

Included LG Camera package in grayscale_icon_map.xml

Some icons do not follow the design standards, however this was done on purpose.


fischer-felix:
Since adding this, it seems DB Navigator has already got an icon, so I removed it from my fork

fischer-felix:
additionally I have made Icons for 
- Cellmapper
- Flightradar24
- Geizhals
- Wayback Machine

fischer-felix:
Is there anything I need to change for this to get merged? I'm asking since the original pr is already two months old and there has not been any activity on reviewing. 
I completely understand id there is just a lot of other more important tasks, but if so it would be nice if you could let people know that so they're not confused as to why nothing gets merged.
Thanks.

patrykmichalik (changes requested):
Hello, and thanks for your interest in contributing! Sorry for the wait. We work on Lawnicons in our free time, so it can be hard to keep up. Some suggestions:

- `microsoft_onenote.svg` doesn’t appear to follow the style of Lawnicons.
- For `mobiles_bezahlen.svg`, I’d suggest increasing the spacing between the lines. We generally ensure that adjacent lines are at least 8 px apart.
- In `notenapp.svg`, the large filled shape in the middle doesn’t appear to match the style of Lawnicons.
- For `solarweb.svg`, I’d suggest increasing the spacing between the lines. We generally ensure that adjacent lines are at least 8 px apart.
- In `vimpay.svg`, the large filled shapes don’t appear to match the style of Lawnicons.
- For `wayback_machine.svg`, I’d suggest increasing the spacing between the lines. We generally ensure that adjacent lines are at least 8 px apart.
- In `wear.svg`, the lines seem to be rather crammed. It may be hard to create more empty space between the letters, so the icon may need to deviate a bit more from the original.
- In `wetter_online.svg`, some of the lines seem to be too thin. We use a width of 12 px in most cases, and a width of 8 px in special cases. However, some of the lines in this icon are thinner than 8 px.

fischer-felix:
I think all the icons should now technically follow the proper style, but I am still not quite happy with wear.svg, would be nice if you could give your opinion on that.
I tried to do it in the style of the ebay icon, however it does not look quite right to me.

patrykmichalik:
Cheers for applying the edits! About `wear.svg`: have you considered deviating from the official icon and using a symbol of a watch with a _W_ inside, for example? I think itʼd be easier to make such a design match the style of Lawnicons.

fischer-felix:
Thanks for the suggestion, will definitely try that

fischer-felix:
Hi, sorry for the long wait, I've been quite busy with school, but now I have more free time again. 

I sketched up a couple of potential candidates for the Wear icon, but would like to have your opinion on them first.

I think the first one looks really nice, but I feel like it only makes sense to have a watch and buds on the icon (since those are the two product categories it is used for)

The latter three make more sense for the app, but there is a lot of blank space in the top right corner.
I don't think I like the rotated watch, but I figured I'd include it just because
One more minor difference is the earbud used (regular galaxy bud/bud pro) but this should not be a dealbreaker, I like both.


wear_watch_only
wear_budspro+watch
wear_buds+watch_rotated
wear_buds+watch

patrykmichalik:
Hello again! No worries about the wait. After all, this is a free and open-source project, not a job. 😆 The first icon looks great to me. I think a watch is sufficient here because it symbolizes a wearable device. In my view, it’s enough to represent the general point of the app. The icon doesn’t have to showcase every device that the app works with.

patrykmichalik (approved):

patrykmichalik:
Thanks, @fischer-felix!

PR #474, opened 2022-05-20
Added WallSweets icon

WaifuPX-DG:
Wallpaper App


patrykmichalik (changes requested):
Thanks for your interest in contributing! For each new icon, an XML file and an SVG file should be added. This is described in [`CONTRIBUTING.md`](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md). However, it would appear that no SVG file is included in this pull request.

PR #475, opened 2022-05-21
Added icons for Tachiyomi & Dealabs

lorrding:
First time contributing, please tell me if anything is wrong.
I don't really know how to draw svgs so I hard drawn them directly from xml


lorrding:
Here's a preview of the icons
lawnicons_tachiyomi_dealabs

PR #501, opened 2022-06-01
Add QKSMS icon

ThatOneCalculator:
Add QKSMS icon

Reference: 

Icon:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Some suggestions:

1. It would appear that the path doesnʼt fill up the recommended content area.
2. In the areas marked with arrows, there appear to be geometric errors. As a result, the curvature of the path is inconsistent.

image

PR #502, opened 2022-06-01
Add Dawn icon

ThatOneCalculator:
Add Dawn icon

Reference: 

Icon:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Iʼd suggest ensuring that the coordinates of all key geometric points are whole numbers. At this artboard size, numbers such as those in the image can generally be avoided. I recommend you turn on the pixel grid in your design tool.

By “key points” I mean points such as the centers of circles, the corners of rectangles, and the vertices of triangles. In the case of boolean operations, the coordinates of the base shapes should be whole numbers, but the points that are created as a result of the operations sometimes have to be non-integers (particularly when you’re working with circles).

image

Goooler:
@ThatOneCalculator Could you please rebase and tweak?

PR #503, opened 2022-06-01
Add Snapdrop

ThatOneCalculator:
Add Snapdrop icon

Reference: 

Icon:


ThatOneCalculator:
Drafting, as making these in Inkscape seems to have caused some issues...

patrykmichalik (changes requested):
Thanks for your interest in contributing! It would appear that the icon is tilted to the right slightly:

image

PR #506, opened 2022-06-01
Add VSCO

ThatOneCalculator:
Add VSCO icon

Reference: 

Icon:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Unfortunately, this icon doesn’t appear to match the style of Lawnicons. The icons in Lawnicons use a uniform stroke width of 12 px.

PR #514, opened 2022-06-02
Add Yet Another Call Blocker

ThatOneCalculator:
Adds Yet Another Call Blocker


patrykmichalik (changes requested):
Thanks for your interest in contributing! For the symbol of the phone, Iʼd suggest using a stroke width of 12 px (the shape may need to be slightly reworked) and ensuring that all key geometric points have whole coordinates. Please see [here](https://github.com/LawnchairLauncher/lawnicons/pull/502#pullrequestreview-1033490033) for more information.

PR #515, opened 2022-06-02
Add Authenticator Pro

ThatOneCalculator:
Add Authenticator Pro


patrykmichalik (changes requested):
Thanks for your interest in contributing! Unfortunately, the icon appears to contain several geometric errors and be asymmetrical.

image

PR #517, opened 2022-06-02
Add Venmo

ThatOneCalculator:


patrykmichalik (changes requested):
Thanks for your interest in contributing! Unfortunately, this icon appears not to match the style of Lawnicons. The icons in Lawnicons use a uniform stroke width of 12 px.

PR #518, opened 2022-06-02
Add MuPDF

ThatOneCalculator:
Add MuPDF


patrykmichalik (changes requested):
Thanks for your interest in contributing! I’d suggest ensuring at least 8 px of room between adjacent lines, ensuring all key geometric points have whole coordinates ([more information](https://github.com/LawnchairLauncher/lawnicons/pull/502#pullrequestreview-1033490033)), and updating the path with a stroke width of 9 px such that the stroke width is either 12 px or 8 px.

PR #520, opened 2022-06-02
Add DeepL

ThatOneCalculator:
Adds DeepL


patrykmichalik (changes requested):
Thanks for your interest in contributing! For the shape inside of the bubble, Iʼd suggest ensuring that the two diagonal lines are 12 px in width.

PR #533, opened 2022-06-03
Add Printers

ThatOneCalculator:
Adds a printer icon to support the most popular printing apps on the Play Store and F-droid, all of which use a very similar icon.


patrykmichalik (changes requested):
Thanks for your interest in contributing! Iʼd suggest ensuring that all key geometric points have whole coordinates. Please see [here](https://github.com/LawnchairLauncher/lawnicons/pull/502#pullrequestreview-1033490033) for more information. Additionally, some lines appear to be slightly tilted.

Chefski (approved):

PR #534, opened 2022-06-03
Add Magic Wormhole

ThatOneCalculator:
Adds support for Magic Wormhole clients


patrykmichalik (changes requested):
Thanks for your interest in contributing! Iʼd suggest ensuring that the icon fills up the recommended content area.

image

PR #537, opened 2022-06-03
Add Manyverse

ThatOneCalculator:
Adds Manyverse


Bnyro:
why one pr for each icon?

ThatOneCalculator:
- I made them all at separate times
- If the maintainers want to merge some but not all, multiple PRs will allow the maintainers to individually review each icon instead of having to make an all-or-none decision

patrykmichalik (changes requested):
Thanks for your interest in contributing! While the original Manyverse icon is rather geometric, comprising four main lines connected with curves that are parts of circles, your icon appears to be somewhat looser. For example, the diagonal lines arenʼt entirely straight. Iʼd suggest bringing the icon closer to the original in this respect.

image

PR #554, opened 2022-06-12
Add plex

lorrding:
lawnicons_plex


lorrding:
Fixes #552

patrykmichalik (changes requested):
Thanks for the PR! Iʼd suggest making the gaps highlighted in the image at least 8 px wide. This way, the gap width would better match the stroke width. Additionally, thin gaps, when viewed at small sizes, can be mistaken for artifacts. Another option would be to extend the two smaller shapes such that there are no gaps at all.

image

lorrding:
Hi, another late fix! I've made the gaps larger tell me if it's still not large enough :)

Chefski (approved):
Looks good! Thanks for making the change! :D

PR #556, opened 2022-06-12
Add credit_mutuel

lorrding:
preview:
lawnicons_credit_mutuel
base logo:
unnamed


patrykmichalik (changes requested):
Thanks for the PR! There appears to be a tiny gap in the center of the icon. Iʼd suggest correcting this.

image

lorrding:
Sorry for the long delay, it should be fixed now 😅

Chefski (changes requested):
Thanks for fixing it, the icons seems to be bigger than the 160x160px content area. Mind fixing that? :)

lorrding:
Hi sorry again for the late fix 😅
I've edited the svg to fit the 160px size! Only thing I can't understand is when exporting to a drawable xml, it change the stroke size to 11.17px ? It may be due to the fact that I'm using a transform matrix to resize the svg 🤷‍♂️

Chefski (approved):
Thanks for making the changes. I fixed the stroke being 11.17

Issue #563, opened 2022-06-14
Icon request


Hiscusaa:
Excuseme can please add icon for this app?
Fkm
Apex Legenda
Tachiyomi/Tachi J2K
![Screenshot_20220615-042709_Lawnchair](https://user-images.githubusercontent.com/107515185/173695218-e005c149-9930-467c-84f8-abb64dbbfcd2.png)

Chefski:
Added all 3 👍

lawnchairci:
Hello! It would appear that this icon request doesnʼt include the package names of some or all of the apps for which icons have been requested. In addition, we are switching to a new system for icon requests. The apps for which package names have been provided (if such apps exist in this icon request) have been added to the new icon request database, meaning no action is required. For the apps for which package names havenʼt been provided, please submit an icon request via the new icon request form: https://forms.gle/Fx8vZAiWdW1Tyjo57.

PR #565, opened 2022-06-15
3 Icons Added

Chefski:
Adds icons from request #563 

Icons added:
- Apex Legends Mobile
- Tachiyon
- Franco Kernel manager


lorrding:
I've already made a Tachi icon here #475  but it was one of the first I made so it's far from perfect😅.
Anyway, J2K is a fork of Tachiyomi so you should link the icon to both apps if we keep your PR.

```xml
<icon drawable="@drawable/tachiyomi" package="eu.kanade.tachiyomi" name="Tachiyomi" />
```

Chefski:
> J2K is a fork of Tachiyomi so you should link the icon to both apps if we keep your PR.

Ahh thanks. I just got the package name from the icon request and didn't really look into it

patrykmichalik (changes requested):
Thanks for the PR! For `franco_kernel_manager.svg`, Iʼd suggest tweaking the letters as follows for greater consistency with the style of Lawnicons.

image

PR #578, opened 2022-06-20
Add Chrome remote desktop

lorrding:
I can make the center Chrome icon thinner if it feels too big


patrykmichalik (changes requested):
Thanks for the PR! Iʼd suggest tweaking the geometry of the icon such that the path is aligned to the pixel grid.

image
image
image

Chefski (approved):
Thanks!

PR #605, opened 2022-07-12
chore: add icon Mt Manager

taamarin:
Screenshot_20220712-092922_Lawnchair


patrykmichalik (changes requested):
Thanks for the PR! Unfortunately, it would appear that the icon contain geometric errors, possibly because it was created using a tool for converting PNGs to SVGs. As a result, the shapes are crooked and the complexity of the path is higher than necessary.

image
image

PR #614, opened 2022-07-21
Oxygen OS 12 dialer app icon is showing as contact icon

bappa555:
Oxygen OS 12 dialer app icon is showing as contact icon


SuperDragonXD:
Thanks for the PR!

Issue #609, opened 2022-07-15
Chrome Icon


ariqren:
The size of the Chrome icon is smaller than the rest.

REM-Codes:
I don't have this issue.

SuperDragonXD:
Could you provide a screenshot or similar?

ariqren:
![1658213976267](https://user-images.githubusercontent.com/85684469/179687214-a354b145-4165-4e0f-9c3e-b5526bcd239d.jpg)

Its not really a big issue actually. Just my OCD acting up. If other people dont have this issue, maybe this bug is only on my phone.

Chefski:
Upon checking this, both Chrome and the Dailer have the correct sizes, however contacts is smaller and messages is off the centre. I'll make a PR for making contacts bigger and bringing messages to the centre (but I think its off because it has a bit sticking out. if its in the centre then it'll look off).
![image](https://user-images.githubusercontent.com/100310118/181631706-293d72b6-4f2a-493f-9109-b9f4cbb64a2e.png)
![image](https://user-images.githubusercontent.com/100310118/181632048-7151fa77-360c-4bf9-bdb6-d7e0280ddea8.png)

Chefski:
This is due to scaling done by the launcher, not the icons themselves.

I'll be closing & marking this issue as a duplicate as it's already been mentioned in #230

PR #620, opened 2022-07-28
Update Messages icon

Chefski:
Messages icon was not meeting icon guidelines
#609 
image


MominRaza:
> Messages icon was not meeting icon guidelines

So you replaced with contact icon 😀

PR #626, opened 2022-08-06
Add AnkiDroid, Nebula and Flux icons

dercodeling:
As the title says, I've added icons for AnkiDroid, Nebula, and Flux, I've successfully tested it on my phone as well.

Android Studio shows errors regarding some contents of the SVG file, though, which seems to stem from the fact that I created them in Inkscape and that Inkscape adds some proprietary attributes and other stuff, that Android Studio says is "not allowed here". Is this a problem or a neglectable? Unfortunately, I don't know how to "clean up" the files if it is problematic, so ideas regarding that are appreciated.


lorrding:
Try to export to ``plain SVG`` instead of ``inkscape SVG``
Then what I do is I use a "blank" svg and grab the ``d`` attribute in each ``<path/>`` with vscode.
Screenshot_1

Here's what I use:
```svg
<svg xmlns="http://www.w3.org/2000/svg" fill="none" stroke="#000000" stroke-width="12"  stroke-linejoin="round" stroke-linecap="round" width="192" height="192" viewBox="0 0 192 192">
	<path d=""/>
</svg>
```

You just need to be carful with the `transform` attribute and the `viewbox`/`stroke-width` you use or the svg will be messed up
Screenshot_2

dercodeling:
I'll open a new PR so I can start from scratch.

PR #657, opened 2022-09-14
Remove `ic_` prefix from all remaining icons

Chefski:
## Description

Removes the `ic_` prefix on icons which still had it

## Type of change


SuperDragonXD (changes requested):
Thanks for your interest in contributing! `ic_lawnchair_background` should not be changed since it is a core component of Lawnicons unthemed adaptive icon (it makes the green-blue background), and isn't really the "icons" part of Lawnicons. Could you re-rename it back?

Chefski:
My bad there. Should've checked the connections to `ic_launcher_background`

SuperDragonXD (approved):
Thanks for applying the changes! Merging the PR.

PR #720, opened 2022-10-01
Add Leboncoin

lorrding:
## Description

Here is a preview :
Screenshot_20221001-221304
 And base icon :
unnamed

## Type of change

### Icons added
* Leboncoin (`fr.leboncoin`)


Chefski (changes requested):
Thanks for another icon addition. I've noticed the icon is a bit smaller bigger than the recommended 144x144px content size for squares. Mind fixing it?

lorrding:
> Thanks for another icon addition. I've noticed the icon is a bit smaller bigger than the recommended 144x144px content size for squares. Mind fixing it?

My bad it's an icon I made a while ago and forgot to check the size 😅
I'll fix this during the week

Chefski (approved):
Thanks for fixing the size :)

PR #726, opened 2022-10-05
Resize vodafone

daywalk3r666:
## Description

### Icons added
* MeinVodafone (`com.appseleration.android.selfcare`)


Chefski (changes requested):
Hey thanks for making the change. I've noticed now that the icon has a stroke with of over 15px. Its caused by the icon have an inner and outer path which both have strokes. Could you reduce the stroke width to 12px?

image

PR #738, opened 2022-10-08
Remove duplicates

daywalk3r666:
## Type of change


Chefski (approved):
Thanks for finding the duplicates!

PR #739, opened 2022-10-08
Add Kotlinter to check Kotlin code style

Goooler:
## Description

https://github.com/jeremymailen/kotlinter-gradle

```editorconfig
[*.{kt,kts}]
ij_kotlin_imports_layout = *
ij_kotlin_allow_trailing_comma = true
ij_kotlin_allow_trailing_comma_on_call_site = true
```

## Type of change


.github/workflows/check.yml
```diff
@@ -0,0 +1,27 @@
+name: Build debug APK
```
SuperDragonXD, line 1: Shouldn't this be changed to "Prettify Kotlin Code", "Check Kotlin Style" or something similar?

Also I think having a more descriptive file name (like `check_kotlin_style`) could be better too.

SuperDragonXD (changes requested):
Just one small suggestion on the `check.yml` file.

.github/workflows/check.yml
Goooler: Ohhhh, I copied from `build_debug_apk.yml`, missed it.

Goooler: f782a64c84270ccdf94525aed0aa23bfb381c2ce

SuperDragonXD (approved):

.github/workflows/check_kotlin_style.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+    branches:
+      - develop
+    path:
+      - '.**/**.kt'
+      - '.**/**.kts'
+  pull_request:
+    path:
+      - '.**/**.kt'
+      - '.**/**.kts'
```
Goooler, line 13: Path filter not work, fixed by bc3966190d27a3310d1d1b65f8e0df060d799fd2

PR #770, opened 2022-10-27
add dealabs icon

lorrding:
## Description

Add dealabs icon. I remade the icon I first made months ago here #475 (I just forgot I closed the pr 😅)

This time I double checked the icon size, no more "144px for squares" fixes needed (hopefully) !

## Type of change

### Icons added
* Dealabs(`com.dealabs.apps.android`)

### DIFF
--- a/app/src/main/res/drawable/dealabs.xml
+++ b/app/src/main/res/drawable/dealabs.xml
@@ -1,0 +1,18 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="192dp"
+    android:height="192dp"
+    android:viewportWidth="192"
+    android:viewportHeight="192">
+  <path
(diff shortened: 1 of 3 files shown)


Chefski (commented):
Hey! The icon unfortunately still appears smaller than 160x160px. I'm not sure if its the same for you but I'm just copying and pasting the svg into Figma. Does it fit the 160x160 box in the software you use?
image

lorrding:
It should be a little bit taller now. I think this might be due to the way I export the svgs from inkscape (`Inkscape SVG` or `plain SVG`)

Also, the base icon is not a perfect square (around 154 by 160 when resized) so it's a little bit taller than wider, I don't know if it makes sense 😅
image

Sorry you still had to point out an issue.

Chefski (approved):

Chefski:
Yea I'm not sure what the issue was there but its good now. Either the width, height or both have to be 160. Thanks for contributing

PR #831, opened 2022-11-16
Link 4 icons

ArtikusHG:
## Type of change

### Icons linked
* UntrackMe (linked `app.fedilab.nitterizeme` to `@drawable/trust_wallet`)
* TimeTable - a FOSS timetable app, icon looks very similar to the news glyph, even though the app has a different purpose (linked `com.asdoi.timetable` to `@drawable/news`)
* Librera Reader - not related to google, but icon looks very similar to google books glyph (linked `com.foobnix.pro.pdf.reader` to `@drawable/google_play_books`)
* Iconify - a FOSS UI customization app, again, totally unrelated to the original glyph but looks very similar (linked `com.drdisagree.iconify` to `@drawable/eksi_sozluk`)

### DIFF
--- a/app/src/main/res/xml/grayscale_icon_map.xml
+++ b/app/src/main/res/xml/grayscale_icon_map.xml
@@ -321,6 +321,7 @@
     <icon drawable="@drawable/ebay_kleinanzeigen" package="com.ebay.kleinanzeigen" name="eBay Kleinanzeigen" />
     <icon drawable="@drawable/eight_ball_pool" package="com.miniclip.eightballpool" name="8 Ball Pool" />
     <icon drawable="@drawable/eksi_sozluk" package="com.eksiteknoloji.eksisozluk" name="ekşi sözlük" />
+    <icon drawable="@drawable/eksi_sozluk" package="com.drdisagree.iconify" name="Iconify" />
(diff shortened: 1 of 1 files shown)


Chefski (changes requested):
Hey! Thanks for the new links. 

Could you change the name of the link to be the same as the the original icon link?

for example:
`<icon drawable="@drawable/news" package="com.asdoi.timetable" name="TimeTable" />`
will change to
`<icon drawable="@drawable/news" package="com.asdoi.timetable" name="News / Nachrichten" />`

This is because links with different names will create duplicates in the icon gallery.

ArtikusHG:
Okay! I should do this for every icon, right?

Side note: wouldn't it be easier (and better) to add a check for duplicate icons in the main menu? This should only take a few lines of code, considering modern languages have built-in methods for removing duplicates in an array. It would also give the user the ability to search for the icon in the app by name.

Just my 5 cents though, if you consider your approach better, I will just change the names.

Chefski:
Best thing here is make the icons for each app. I don't know Kotlin to make it not include duplicate links with different names. You can add a ` / ` between the names of you want like in Just Eat.

ArtikusHG:
Here you go, I changed the icon names to be like the originals, as you requested.

PR #840, opened 2022-11-18
Added Joyn Icon

SpaceFrrog:
### Icons added
* Joyn (`de.prosiebensat1digital.seventv`) - added from icon request form

It is my first Pull Request, so let me know, if something I should change.


Chefski (changes requested):
Hey! Thanks for contributing to Lawnicons! Could you make the icon bigger by increasing it content area size to 160x160px? You can also make the stroke 12px as letters will be bigger.
image

SpaceFrrog:
Now the size and stroke width should be right.

Chefski (approved):
Thanks

PR #847, opened 2022-11-21
Add icon for xvii and currencies apps

ArtikusHG:
## Type of change

:white_check_mark:  Icon addition (non-breaking change that adds/modifies Lawnicons's icons)

### Icons added
* xvii (`com.twoeightnine.root.xvii`) - a FOSS app for VK messenger. The icon of the original app is literally just a cross, so mine is the same, but following the lawnicons style guidelines.
* Currencies (`de.salomax.currencies`) - a FOSS currency converter. My icon resembles the logo of the original.


Chefski (changes requested):
Hey! Could you increase the icon to fit the 160x160px content area size? Can you also reduce the stroke to 12px and try keep it stroke instead of fill?
image

ArtikusHG:
Hey, in my opinion, 160x160 is a bit large for this specific icon, but I will do it tomorrow if I absolutely must. Also, sorry if I did anything wrong while rendering, I have no idea what I'm doing in InkScape

ArtikusHG:
Here you go, it should now be exactly in the 160x160 area.

Chefski (approved):
Thanks. Completely understand that some icons feel like they're too big but if we make them smaller than each icon would look weird with different sizes

PR #848, opened 2022-11-21
Add bol

daywalk3r666:
## Description

https://play.google.com/store/apps/details?id=com.bol.shop

✅ Icon addition (non-breaking change that adds/modifies Lawnicons's icons)

### Icons added
* App Name (`com.bol.shop`)


Chefski (changes requested):
Hey! Could you decrease the icon a bit as it goes outside the content area size. You could also make it less tall as the original icon isn't as tall.
image

daywalk3r666:
I will make icon new

PR #854, opened 2022-11-25
Add icon for openScale

Sir-Photch:
## Description

I noticed that the icon for openScale is missing from lawnicons. I tried to come up with a svg. I hope that it fits the requirements.

## Type of change

:white_check_mark: : Icon addition (non-breaking change that adds/modifies Lawnicons's icons)

### Icons added
* App Name (`com.health.openscale`)

### Icons linked
* App Name (linked `com.health.openscale` to `@drawable/openscale`)


Chefski (changes requested):
Hey! The circle does not seem to be perfectly round and the exporting seems to be a bit a strange. Did you export it with a height and width of 192x192?

Could you make the circle round?

Sir-Photch:
The width and height should now be properly set to 192x192.
I also adjusted the circle. Let me know if there is anything else

Chefski (changes requested):
Thanks for fixing the width and height. Unfortunately there are a few changes needed:

- The circle is still not perfectly round
- The icon goes outside the 160x160px content area size. It needs to be made a bit smaller
- The shape inside does not join with the other lines which leaves a gap between both lines

image

Could you also round this line?
image

Sir-Photch:
Here goes my third attempt

Chefski (changes requested):
Cool. Last two things I think. Not sure if I only noticed it now but the strokes have an opacity of 99.8% instead of 100%. Also the circle doesnt have a stroke of 12px.

Sir-Photch:
I double checked about the stroke width now. By chance, do you know if using Inkscape and exporting it as an "optimized" svg causes issues?

Chefski (approved):
Looks good. Thanks

Chefski:
I don't know if Inkscape has any issues. Used it maybe once so I basically have no experience with it

PR #910, opened 2022-12-21
Add 4 icons

peterge1998:
## Description

Fixes #(issue)

## Type of change

### Icons added
* BeReal
* Tabletop Tools
* imgur
* waistline


Chefski (changes requested):
`bereal.svg` does not have whole numbers for it's strokes. Could you change it to 8px (I'll try make different guidelines for letters). If you dont like the look of it in 8px, you can maybe use the favicon on [bere.al](https://bere.al/en).

`imgur.svg` could use 12px. Was the decision to use 8px to make it closer to the original icon?

peterge1998:
> `bereal.svg` does not have whole numbers for it's strokes. Could you change it to 8px (I'll try make different guidelines for letters). If you dont like the look of it in 8px, you can maybe use the favicon on [bere.al](https://bere.al/en).

What do you mean with using the favicon? Just B.?
The line numbers are fucked up because it looks like shit otherwise. In the a I need like two width for this font...
 
> `imgur.svg` could use 12px. Was the decision to use 8px to make it closer to the original icon?

Done

Chefski:
> What do you mean with using the favicon? Just B.? The line numbers are fucked up because it looks like shit otherwise. In the a I need like two width transitioning to each other for this font...

Would a stroke of 6px work better? Do you think it would be suitable as a global stroke width for all text?

peterge1998:
> Would a stroke of 6px work better? Do you think it would be suitable as a global stroke width for all text?

I prefer B. ...

Chefski:
Sure! You can make the B.

peterge1998:
> Sure! You can make the B.

Done :)

Chefski (approved):
Thanks

PR #884, opened 2022-12-04
Lawnicons as icon pack (cleanups)

naveenccmsd:
This PR can be merged after https://github.com/LawnchairLauncher/lawnicons/pull/815 .

PR #815, opened 2022-11-10
Lawnicons as icon pack (core changes)

naveenccmsd:
## Description

- LawnIcons as icon pack
- Two apk default one (icons are dark ) and another apk with light icons(white lines)
- Introduce automation to reduce some maintance work. 
      - Since lawncions will act as icon pack, we need to add components details to appfilter.xml , drawabale.xml and grayscale_icon_map.xml .  Even now developers are creating svg  files and vector assests.
      - Idea is to let developer create svg file and appfilter.xml file alone and helper script to create others file during build.

Demo:

https://user-images.githubusercontent.com/7315975/201498198-fad3c058-89b6-4510-802b-1f560ee8bd76.mp4

Issues:
I see svgs and vector asset is not in sync since they added or created manually. They both need to be in sync inorder for lawnicons to work.

Action Items:

- [x] Update configs to recognise lawicons as icon pack
- [x] Update workflow to build 2 apk 
- [x] Create helper class to create xml and vector automatically
- [x] create/update svg as per vector asset  (noticed some svgs width and height updated in vector asset ). 
- [x] Update Activity for Some apps [app/assets/appfilter_notMapped.xml](https://github.com/LawnchairLauncher/lawnicons/pull/815/files#diff-78d8683a60dc7b78b04e8837085b82c04e2ebf1d5725958b4e6b9bd4d99fedb0)

Previous Icon mapping : **1098**
Current Icon Mapping : **2472**
Pending to update App Activty : **79** (Added activty to all apps avalible in playstore except paid ones)

PR #988, opened 2023-01-23
Relinking some apps

Sir-Photch:
## Description

It seems like some links got lost while merging #815 and #884.
I relinked some of the apps that caught my attention, but I am sure there are others left.

Also, it might be useful to state in CONTRIBUTING that the list is supposed to be sorted by `name`? This was not quite clear to me in the beginning. Maybe it would make sense to put the name-tag as the first tag in each xelement?


Chefski (approved):
Some apps will not be linked properly now unfortunately... And I'm not sure if the name can be at the beginning

Sir-Photch:
> Some apps will not be linked properly now unfortunately...

Why is that?

> And I'm not sure if the name can be at the beginning

I remember from my C#-days that XAttributes in XElements can just be accessed via the name directly. So something like:
```csharp
var nameAttr = xElement.GetAttribute("name");
```
should work, rendering the order of attributes irrelevant. I have no experience with Kotlin though so your mileage may vary :)

Chefski:
> Why is that?

Before the #815 merge, we didn't add app activity to the links so I think the new links are just added in an automated way. For example the OnePlus Photos had 4 links or maybe even 5 but none actually linked to my Photos app.


> should work, rendering the order of attributes irrelevant. I have no experience with Kotlin though so your mileage may vary :)

I'll ask a team member about this

PR #996, opened 2023-01-26
Enhancement on icon pack 

naveenccmsd:
## Description
- Removed two apk (light and dark)
- Icon color will have system accent color by defult.
- Added background to icons (previously it was without background and icon sizes are bit larger if no themed icons selected).


daywalk3r666:
Hey. Can you maybe help me to adapt the svg processor to my fork?

naveenccmsd:
Sure , Isn't same as lawnicons? Are you facing any issue??

daywalk3r666:
I use the old lawnicons base.  I have no idea what i must change in the code.  I try to adapt it many times but the workflow failed

SuperDragonXD (changes requested):
Please change the workflow file as well, and also the python file.

Also I think we can safely return back to the old Lawnicons resource naming scheme (without the light/dark prefix, i.e. from `res/dark/drawable` to `res/drawable/`), since it's all inside the single app. If that's not possible, maybe rename the `dark` folder to `icons` or something.

naveenccmsd:
> Please change the workflow file as well, and also the python file.
> 
> Also I think we can safely return back to the old Lawnicons resource naming scheme (without the light/dark prefix, i.e. from `res/dark/drawable` to `res/drawable/`), since it's all inside the single app. If that's not possible, maybe rename the `dark` folder to `icons` or something.

The only challenge to use old resources is we are having grayscale_icon_mapping.xml in actual resource file for the icon for which we don't have app components details so they will still applied in old style. I will rename folder to `icons` .

app/build.gradle.kts
```diff
@@ -63,16 +63,13 @@ android {
 
     flavorDimensions += "product"
     productFlavors {
-        create("dark") {
+        create("app") {
```
naveenccmsd, line 66: have used name app, let me know if we can use other variant name like `default` or anyother

SuperDragonXD: It's fine.

SuperDragonXD (approved):

PR #1020, opened 2023-02-01
Add a workflow to automatically optimize SVGs

Chefski:
## Description

SVGs will now be optimized/minified every week. This will remove the need to manually optimize.

## Type of change


Chefski:
Tested it in a private repo and worked fine when I ran the workflow manually.

.github/workflows/optimize_svgs.yml
```diff
@@ -0,0 +1,25 @@
...
+  schedule:
+    - cron: "0 0 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  svgs:
+    name: SVGs
```
Goooler, line 13:
```suggestion
```

Goooler: Just keep all job names lowercase.

.github/workflows/optimize_svgs.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Optimize SVGs
+        uses: ericcornelissen/svgo-action@v3
+        id: svgo
+      - name: Commit optimizations
+        uses: stefanzweifel/git-auto-commit-action@v4
```
Goooler, line 22: If we can configure @lawnchair-bot to commit this? CC @yasanglass

.github/workflows/optimize_svgs.yml
```diff
@@ -0,0 +1,25 @@
...
+        if: ${{steps.svgo.outputs.DID_OPTIMIZE}}
+        with:
+          commit_message: Optimize ${{steps.svgo.outputs.OPTIMIZED_COUNT}} SVG(s)
```
Goooler, line 25: If we can skip commiting if `OPTIMIZED_COUNT` is 0?

.github/workflows/optimize_svgs.yml
Goooler, line 13:
```suggestion
  optimize-svgs:
```

.github/workflows/optimize_svgs.yml
yasanglass: Probably, but I can not look into it these days.

.github/workflows/optimize_svgs.yml
Chefski: Updated the version now. Previously it would keep making the same commit each time the workflow ran by trying to optimise an icon which has already been optimise. Now, if there are no SVGs to optimise, there won't be any commit.

SuperDragonXD (approved):

PR #1081, opened 2023-03-01
Replace PeduliLindungi with SATUSEHAT

ZhmieXD:
## Description

The Indonesian Ministry of Health recently rebranded their COVID-19 Tracer app PeduliLindungi to SATUSEHAT and a new logo to go with it. The rebrand also appears to be a part to turn the app into not just a COVID tracer app, but also a unified place for other medical purposes, like checking the user's general medical record, as well as their policy status on the ministry's nationwide medical insurance program (BPJS Kesehatan). So I figured I should update the lawnicon for that to reflect the change 😅

Also is adding a vector asset for a new icon no longer necessary now? I found out that the XMLs in the drawable folder no longer exist and the contributing guide to add vector assets is no longer mentioned.

In any event, thanks once again for the chance! 😁 

## Type of change

### Icons updated
* PeduliLindungi `@drawable/pedulilindungi` ->> renamed to SATUSEHAT `@drawable/satusehat` (`com.telkom.tracencare`)


Chefski (approved):
Thanks. Vector assets are no longer needed

ZhmieXD:
> Thanks. Vector assets are no longer needed

That's time-saving then, especially when adding icons on bulk. Thanks!

PR #1141, opened 2023-04-02
adding Alipay app as ali_pay.svg

medtty:


projjalm:
The added svg looks very similar to alipay.svg, which I think it is meant to replace.

projjalm (commented):
The added svg looks very similar to `alipay.svg`, which I think it is meant to replace.

Chefski (commented):
Icon is the same as `alipay.svg` except this one isn't linked to anything. This seems unnecessary the Alipay already has an icon

Chefski:
I'll be closing this PR as it adds the same icon but with a different file name. Thanks for contributing

PR #1142, opened 2023-04-02
Adding ViMusic app as vi_music.svg

medtty:


projjalm (commented):
I think this svg looks better than mine for the same app (ViMusic) which got merged sometime back (#1112). It just needs a bit of scaling and realigning to fit into the 160x160px grid centrically.

Chefski:
image
Is this the original icon?

projjalm:
Yes, that one.

projjalm:
Ah, no, if you find the previous one more suitable then let it be that way :)
The reason I thought this one looks better is because of its simplicity, but reviewing it again now makes me think it is, way too simple.
Let the previous icon remain and wait for @medtty to comment.

Chefski:
I'll be closing this PR. The current icon has been agreed upon as more suitable. Thanks for contributing

PR #1149, opened 2023-04-08
icontool.py guide

### LINKED ISSUE #1134: [DISCUSSION] Document icontool.py in the contributing guide
### About this discussion

A tutorial/guide for using icontool.py should be added to CONTRIBUTING.md
I want to help but i'm not good at writing documentation

Chefski:
## Description
Adds a guide/documentation on how to use `icontool.py`.

Fixes #1134 

## Type of change


Chefski:
Wow my commits are messy...

thismoon:
Thank you for making a documentation!

can you change the 47th line of `icontool_guide.md` to "## Remove a link and *its* SVG"

also instead of having "Add an icon" and "Add an icon (with generated message)" maybe it's better to keep the first one and put a small notice below it that an `-m` argument can be added to print a generated message to be used in a pr

and shouldn't code blocks be in `sh` instead of `xml` since these are commands?

Chefski:
> can you change the 47th line of icontool_guide.md to "## Remove a link and its SVG"

Done

> also instead of having "Add an icon" and "Add an icon (with generated message)" maybe it's better to keep the first one and put a small notice below it that an -m argument can be added to print a generated message to be used in a pr

How do you think we could do this? I just looked at other documentation and they generally did something like this. Just a small message underneath to add `-m` at the end for a message?

> and shouldn't code blocks be in sh instead of xml since these are commands?

I went with xml cuz its also used in `CONTRIBUTING.md`. I think it's just better since it keeps all the colours monotone.

thismoon:
> How do you think we could do this?

you can add a note hightlight
> **Note**
> an `-m` argument can be added to print an optional generated message

the problem that it should be in a blockquote, also i didn't see this method used in this repo
or just add something like this:

**Note:** an `-m` argument can be added ...

Chefski:
I added it as a note. What you think now?

thismoon:
Good, It's clean now. I think it's ready to be merged

PR #1163, opened 2023-04-13
Add Icons

Vitaliklidy:
### Icons added
* HIT (`com.lionwood.nitwrap`)
### Icons Linked
* My Vodafone (`ua.vodafone.myvodafone`)


Chefski (changes requested):
Hey! Thanks for contributing.

The letters in `hit.svg` have a thickness of around 10/11px. Could you make it either 8 or 12px? Could you also round the corners.
image

For `my_vodafone.svg`, I'm wondering if `vodafone.svg` already present in the svgs folder can be used instead. It seems like the same icom

Vitaliklidy:
Yes, the My Vodafone and Vodafone has same icon

Chefski:
why not just link component info to the already existing icon then instead?
vodafone

Vitaliklidy:
Ok I link it

Vitaliklidy:
I change hit icon is it better ?

PR #1164, opened 2023-04-13
Change minecraft logo to make look more like original

Vitaliklidy:
### Icons Changed
* Minecraft (`com.mojang.minecraftpe`)


Chefski (changes requested):
This icon has more of a fill style to it rather than outline/line. Corners should be rounded too. Maybe some kind of outline design could be applied to it and we can see what it looks like?

Grabstertv:
The new design has user experience issues. Its making it hard to recognize as a minecraft icon.

PR #1178, opened 2023-04-16
Add Meetup

Sir-Photch:
## Description

Adds icon for Meetup

## Type of change

:x: Bug fix (non-breaking change which fixes an issue)
:white_check_mark: Icon addition (non-breaking change that adds/modifies Lawnicons's icons)
:x: General change (non-breaking change that doesn't fit the above categories like copyediting)

### Icons added
* App Name (`com.meetup`)

PR #1182, opened 2023-04-18
Add meetup

Sir-Photch:
Compare #1178


Chefski (changes requested):
As previously mentioned, could you change the icon so that the two paths in the `m` don't join a make one thick stroke. With the new guidelines you can use 6px if required however I'd recommend trying to maybe get rid of the dots around it or something else. Also at the top left or the bubble the path isn't quite smooth. You can look at the previous screenshot.

Sir-Photch:
Will do eventually, though which screenshots do you mean?

Chefski:
> though which screenshots do you mean?

In this PR #1178

Sir-Photch:
Is it just me or aren't there any comments with screenshots?

Chefski:
Could be just you:
image

Sir-Photch:
I've tried removing the dots and increasing the size of the icon, but then the two lines of the 'm' still connect. So I changed it to 6px. Also, I've smoothed the edge you were mentioning.

Chefski (approved):
Thanks

PR #1185, opened 2023-04-18
Fix #1184

### LINKED ISSUE #1184: Home Assistant icon not working
### About this discussion

Home Assistant is in the appfilter.xml, however when I add it to my home screen it doesn't get a custom icon. Everything else works. Any reason why?

Artem13327:
I just copied the line above and pasted the value from the issue creator into it. The string I added is probably for an old version of the application, since I had no problems with the icon. This comment was written through a translator, sorry for any mistakes
Before accepting, please seek approval from @KTibow. Since I myself did not check this Pull Request


KTibow:
hey, I don't think this will work, but for the sake of testing how would I build an APK?

Chefski:
@KTibow download the debug file from [the actions tab](https://github.com/LawnchairLauncher/lawnicons/actions/runs/4735629621).

KTibow:
actually doesnt help, as that's not the launcher activity, and i just had an outdated version
however if you could make the ha icon use the less detailed version that would be appreciated - see https://github.com/home-assistant/assets/blob/master/logo/logo-small-black.svg
or at least fill in the holes as the current version is way too detailed

Chefski:
So is the icon working as intended? @KTibow

PR #1186, opened 2023-04-19
Add Hay Day icon

billy4479:
## Description

Adding the Hay Day icon.

## Type of change

### Icons added
* Hay Day (`com.supercell.HayDay`)

### Icons updated
None

### Icons linked
None

***

Reference:

Traced icon:
Hayday

As you can see, the reference icon cuts the crest of the chicken. I completed it by going over the 160px of the icon's content, I'm not sure what is the procedure in this case but I can resize it if needed.

The path stroke it's all 6px wide because of the amount of details in the icon. If a wider stroke is required I can try to modify it but it does clutter the icon a bit.

Also, I've noticed that `app/assets/appfilter.xml` has CRLF line endings. If this is the intended behavior then `icontool.py` should also enforce it on every system, since I was adding this icon on Linux and got a huge diff because it converted all line endings to LF. I fixed it using `unix2dos` but I think a default should be enforced maybe with a git hook on every commit or some other solution.

That said, I hope I did everything correctly, if there's any problem let me know.


Chefski (commented):
I think 8px would work with this icon

billy4479:
I'm not sure why CI is failing.. It probably doesn't like the fact that I rebased and force pushed my branch. I can open a new PR if that's a problem.

SuperDragonXD:
> I'm not sure why CI is failing.. It probably doesn't like the fact that I rebased and force pushed my branch. I can open a new PR if that's a problem.

Yes, force pushing doesn't really make sense for the checks: https://github.com/LawnchairLauncher/lawnicons/actions/runs/4744556196/jobs/8431799467#step:6:1

Could probably add a `try ... except git.GitCommandError` block so that the checks will still work

svgs/hay_day.svg
Chefski, line 1: Could you just fix the icon to be inside the content area
image

Chefski (changes requested):

svgs/hay_day.svg
billy4479: Which should be 160x160 or 144x144?

Chefski: 160x160

Chefski (approved):
Thanks

PR #1181, opened 2023-04-18
Add 37 icons

KevoM:
Fixing conflict

PR #1193, opened 2023-04-22
Update George Icon

vascYT:
## Description
I noticed that I included the wrong app filter for an app icon I submitted 10 days ago. I fixed it with this pull request, so the icon works now.
### Icons updated
* George Austria/Österreich (`at.erstebank.george`) - updated app filter


Chefski:
Hey! Seems like you may have the sae issue as [here](https://github.com/LawnchairLauncher/lawnicons/pull/1181#issuecomment-1518362439). Could you try use unix2dos or dos2unix as mentioned in the reply

vascYT:
I tried everything. The file I commited uses LF line endings. And your .gitattributes file also configures it to use LF endings, so it gets automatically converted. I don't know why it is still not working...

vascYT:
I noticed that your current file in the develop branch is using the CRLF line terminators:
image
But your .gitattributes states that all text files should use LF terminators:
image
So for every commit it changes the file to LF (for me at least):
image
That's why I am unable to fix this.

vascYT:
I think I found the commit that changes the line endings to CRLF: https://github.com/LawnchairLauncher/lawnicons/commit/d326d05824ee2b32218ce4919fee515120205606

SuperDragonXD:
I'll make a new commit switching the line endings in `appfilter.xml` so it's consistent with `.gitattributes`.

vascYT:
Perfect, thanks!

Issue #1217, opened 2023-05-05
[DISCUSSION] icontool.py keeps the original path in appfilter.xml after adding a new icon


thekeyboardworrier:
### About this discussion

Icontool is leaving full path to the icon instead of it's drawable name in `svgs` folder.

I added an icon with this command
```
python .\icontool.py add C:\Users\Khalid\Desktop\banglalink.svg com.arena.banglalinkmela.app/com.arena.banglalinkmela.app.ui.splash.SplashActivity MyBL
```

And it added this line in the `appfilter.xml` file
```xml
<item component="ComponentInfo{com.arena.banglalinkmela.app/com.arena.banglalinkmela.app.ui.splash.SplashActivity}" drawable="C:\Users\Khalid\Desktop\banglalink" name="MyBL" />
```

It left full path to the icon without the file extension. Expected result was `drawable="banglalink"`

PR #1219, opened 2023-05-05
Icontool fix for underscore in packagename and drawable name for adding icons

### LINKED ISSUE #1216: [DISCUSSION] `icontool.py` cannot parse package name or activity name containing underscore
### About this discussion

I tried adding an app that had underscore in it's package name, it gave me an error saying my formatting was wrong. It wasn't. I added it manually.

Later I looked into the `icontool.py` and found that the RegEx for [PACKAGE_NAME]/[APP_ACIVITY_NAME] parsing does not account for any underscore. Only one line of code change should fix it.

thekeyboardworrier:
## Description
Added fix for `icontool.py` not being able to parse package name or activity name with underscore and it leaving full path instead of basename for svg drawable name for add icon action.

Fixes #1216 and #1217 

## Type of change

:white_check_mark: Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/icontool.py
+++ b/icontool.py
@@ -5,6 +5,9 @@ import shutil
 import os
 import re
 
+# see https://regex101.com/r/xC9Kh3/1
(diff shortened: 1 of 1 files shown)


SuperDragonXD (approved):
Thanks! Looks like I overlooked some parts of the regex.

PR #1232, opened 2023-05-13
Add 1 icon

thekeyboardworrier:
## Description
Add 1 icon and link 2 entry of Toffee app.

Don't freak out by the amount of changes in `appfilter.xml`. It's formatting and `item` entry reordering *mostly* done by `icontool.py`.

## Type of change

### Icons added
* Toffee (`com.banglalink.toffee`)

### Icons linked
* Toffee (linked `com.banglalink.toffee` to `@drawable/toffee`)


svgs/toffee.svg
Chefski, line 1: Hey! Could you try maybe preventing the paths from touching each other? If possible, could you also make this icon in use strokes rather than fill. It's easier to review and will make it easier edit in the future.

Chefski (changes requested):

svgs/toffee.svg
thekeyboardworrier: > Hey! Could you try maybe preventing the paths from touching each other? If possible, could you also make this icon in use strokes rather than fill. It's easier to review and will make it easier edit in the future.

I can use strokes, but I'm nor sure if I am able to prevent the strokes from touching each other. Because the paths in the original logo is almost touching at the top. Even 6 px strokes will touch each other. Here is the monochrome version of the logo, with 6px strokes and with 8px strokes side by side.

toffee_monochrome toffee_6px toffee_8px

Let me know which one you prefer, 6px or 8px version. I'm using 8px one with stokes (instead of path) for now.

Thank you.

Chefski: The shapes can just be reduced in size. So instead of being on the outline of the tear drop for example, the centre of the path will be inside the shape

thekeyboardworrier: Changed it again. But this one has paths instead of strokes. It's easier to keep the shape of the logo. I can't keep the corners sharp with strokes. It gets more blobby.

Chefski: Would it not be better to just make the shapes smaller? I'm not sure in what way this one was made but the previous ones were more suitable for Lawnicons

thekeyboardworrier: Changed it again. Thank you.

Chefski: Looks good!

Chefski (approved):

PR #1243, opened 2023-05-18
Improved text, rounded images corners

RaduBratan:
## Description

I fixed some syntax, punctuation and vocabulary errors while also rounding the corners of the first, third and fourth files in the `images` folder (previously, only the second image had rounded corners). The corners are ~~8px~~ 6px in radius.

## Type of change


RaduBratan:
Actually, now that I think about it, is the radius of the contributing-image-2's corners equal to 8px or 6px? Considering that the icon guidelines mention that corners should be 6px, does this image follow the same rules?

RaduBratan:
Hmm, it still says it has conflicting files, yet I just updated the image to fit the new official changes.

Chefski (approved):

PR #1248, opened 2023-05-19
Added 1 icon, changed 2

RaduBratan:
## Description
I added an icon for MIUI Hidden Settings and changed the other two settings-related icons (MicroG Settings and Android Settings) because they were asymmetrical, not properly sized, had a lot of stray paths and anchor points, some anchor points that weren't stray weren't properly connected and the shape itself was filled instead of being a path with 12px stroke. The side effect of this change is, however, that the overall shape/silhouette of the icon is slightly different.

## Type of change

### Icons added
* Hidden Settings For MIUI (`com.ceyhan.sets`)
### Icons updated
* Settings (`com.android.settings`)
* Vanced microG Settings (`com.mgoogle.android.gms`)


RaduBratan:
The thin stroke of the second cog from hidden_settings.svg looks bad, I'll have to update it.

svgs/hidden_settings.svg
Chefski, line 1: The icon could be increased in size slightly and centered

Chefski (changes requested):

svgs/hidden_settings.svg
RaduBratan: Actually, all I had to do was to move the smaller cog 1px to the right and 1 to the top. While creating the icon, that little cog used a 12px stroke, and the shape was centerd properly. Later on I changed it to 10 px because it looked too thick, but forgot to update it's position.

Chefski (approved):
Thanks

svgs/hidden_settings.svg
Chefski: Ahh no problem :)

PR #1267, opened 2023-05-27
Add C24

sal0max:
## Type of change

### Icons added
* C24 Bank (`de.c24.bankapp`)


svgs/c24.svg
Chefski, line 1: Could you change the background to be transparent please?

Chefski (changes requested):

svgs/c24.svg
sal0max: Sorry! Fixed it now

Chefski (approved):
The SVG has a lot of groups which make the size bigger. If you could ungroup them all at the end it would drastically reduce file size. I went from 5 KB to 947 bytes by ungrouping everything

Issue #1318, opened 2023-06-28
[DISCUSSION] Automatically remind contributors about Lawnicon guidelines in the first PRs


x9136:
### About this discussion

Most reviews to PRs say where the icons don't match [Lawnicon guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md#icon-guidelines). This means that at first the contributors are ill-informed about the guidelines and don't understand the usefulness. It would be helpful to provide a link to the guide after creating a PR with svg changes. Repeat this first 3 PRs and explain what it's for. It should be done by a bot. It's should be meaningful and relatively short.

**Example**
"Make sure the size of the icons, line thicknesses, and corner radius match the [Lawnicon guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md#icon-guidelines). Correct if anything is wrong and update the PR. If everything is according to the guidelines, the reviewers will quickly accept the PR. If something is wrong, you will be asked to fix the icons."

SuperDragonXD:
Could make a GitHub action for the first message. Maybe it could be:

> Thanks for your interest in contributing to Lawnicons!
> 
> Ensure that the size of your icons, their line thicknesses, and their corner radius match the [Lawnicons guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md#icon-guidelines).
> If something is wrong with your icons, please correct them and update the PR.
> 
> If everything is according to the guidelines, the reviewers will quickly accept the PR. Otherwise, you will be asked to fix the icons.

Regarding the above message: could work, but I'm not sure that all contributors use `icontool.py`. 

Also, I kinda like your idea of pre-checking the svgs, either when adding (auto checks) or manually like this:
```console
$ ./icontool add /path/bad.svg example.app/example.app.activity "App Name"
checking if bad.svg is following lawnicons guidelines

* `stroke-color` value is not black `#000000`

error: failed svg checks. use --force to bypass checks (not recommended)
$ ./icontool add /path/bad.svg example.app/example.app.activity "App Name" --force
warning: bypassing svg checks. be sure to provide a reason if someone reviews your svgs.
added App Name to `@drawable/bad` in line ## of appfilter.xml

successfully completed tasks
$ ./icontool add /path/ok.svg example2.app/example2.app.activity "App2 Name"
checking if ok.svg is following lawnicons guidelines
[...]

PR #1330, opened 2023-07-03
add workflow file to check SVGs in PR

lorrding:
## Description

As discused in #1318, I tried to create a PoC of a workflow that could help reviewers quickly see common mistakes made by contributors as well as helping said contributors find out what need to be changed.

You can test the workflow on a repo I made [here](https://github.com/lorrding/workflow-test). I already made 2 PRs with and without issues. Feel free to create any PR with .svg files to test out and see if everything works as intended.
Also, feel free to suggest modifications that could be made to the comment building part or suggest other rules I forgot to include.

This is my first time digging into workflows so this version is far from optimized.

## Type of change

✅  General change (non-breaking change that doesn't fit the above categories, such as copyediting)

### DIFF
--- a/.github/workflows/check_pr_svgs.yml
+++ b/.github/workflows/check_pr_svgs.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Create issue file
         run: |
-          echo "# Some of the svgs don't conform to the Lawnicon guidelines. Fix the svgs so that reviewers would have less to check and clarify deficiencies." > issues.json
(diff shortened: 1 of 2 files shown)


lorrding:
Hi, I've made a lot of progress!

This workflow should now work, you can still test it [here](https://github.com/lorrding/workflow-test).

It's far from perfect and may need some discussion and rework especially on the message part.
I added `svg-check.js` this file is used to run the verification of the .svg files fed to it. It then outputs a .md file with every found issues. The file can be run independently from the workflow like so:
```sh
node svg-check.js {file-to-check.svg file-to-check.svg file-to-check.svg}
```

It's the easiest way I found to make this work properly.
Suggestion and feedback are welcome!

PR #734, opened 2022-10-07
Fix sizing issues from my previous PR / Clean Map a little

leifniem:
## Description

In my previous PR some of the icons were too small, which should be fixed by now.
I redid Adaway and Infinity to be a little more detailed and up to spec, i hope that's okay.

### Icons added
* Post & DHL (`de.dhl.paket`)
* Enpass (`io.enpass.app`)
* Feedly (`com.devhd.feedly`)
* Figma (`com.figma.mirror`)
* Geometric Weather (`wangdaye.com.geometricweather`)
* Loop Habit Tracker (`org.isoron.uhabits`)
* Notion (`notion.id`)
* Obsidian (`md.obsidian`)
* SeriesGuide (`com.battlelancer.seriesguide`)
* THENX (`com.sysops.thenx`)

### Icons updated
* AdAway (`org.adaway`)
* Infinity for Reddit (`package.name`)

### Icons linked
* LMC GCam (`com.google.android.GoogleCameraLMC`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`de.lieferservice.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`lu.takeaway.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`nz.co.menulog.m`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`pt.takeaway.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`ro.ascendnet.android.agregator`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`uk.takeaway.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`at.lieferservice.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`be.pizza.android`)
* Just Eat / Pyszne.pl / Lieferando.de / Takeaway.com / Thuisbezorgd.nl / GrubHub (`ch.lieferservice.android`)
[...]

PR #1340, opened 2023-07-08
deprecate Obsidian icon

Mustang-ssc:
## Description

md.obsidian custom icon can be removed due to:
1. app logo re-design.
2. app includes their own themed icon which is overwritten by this custom icon.

This change reverts md.obsidian icon addition originally added from https://github.com/LawnchairLauncher/lawnicons/pull/734/commits/78bf188b5cf603380c735c8a33234f36dab4b216

## Type of change

:heavy_check_mark: General change (non-breaking change that doesn't fit the above categories, such as copyediting)

### Icons updated
* Obsidian (`md.obsidian`)


Chefski:
I think the icon can stay until someone redesigns it. Better to have an icon for an app than not in my opinion

Mustang-ssc:
> I think the icon can stay until someone redesigns it. Better to have an icon for an app than not in my opinion

I agree, however now the app includes its own adaptive icon unlike previously.

From my understanding this icon is overwriting the app's one which was recently themed by the dev.

The objective of this PR is to let this new adaptive icon plus it's new logo design be shown instead of the custom one.
On the other hand if this may negatively affect older android versions feel free to close it.

RaduBratan:
I'm not a mod/admin/owner or anything, but here's my two cents. I believe the purpose of Lawnicons is to create an unified and consistent icon pack with it's own rules. If an app developer creates their own adaptive icon, that won't necessarily mean it will fit this icon pack, and it will break the consistent look.

Therefore, I would suggest to keep this icon until someone designs a new version (one created specifically to fit this icon pack).

Chefski:
I'll be closing this PR. As stated by @RaduBratan, Lawnicons isn't meant to apps that don't yet have an adaptive icon. If I'm not wrong, many apps have adaptive icons but we still have an icon for them in Lawnicons. Thanks for your interest to contribute!

PR #1399, opened 2023-08-16
Add Squawker

thismoon:
## Description

Added an icon for Squawker

## Type of change

### Icons added
* Squawker (`org.ca.squawker/org.ca.squawker.MainActivity`)


svgs/squawker.svg
x9136, line 1: The eye has a fill, but you need a stroke. You can use a 6x6px circle with a stroke thickness of 6px. I would probably bring the height of the shape to 148px (currently 146.29px).

x9136 (changes requested):
Please fix the bird's eye and resolve conflicts.

svgs/squawker.svg
thismoon: > You can use a 6x6px circle with a stroke thickness of 6px

3x3px circle would work since the program i use have strokes in the center only and making a 6x6px one will leave space inside

> bring the height of the shape to 148px (currently 146.29px).

are you considering the stroke as a part of the shape? inkscape (and the contributing guide) does

inkscape
guide

i could find that number (146) when using figma (which doesn't add the stroke to the height)

figma

I made some changes, is [this](https://cdn.discordapp.com/attachments/842838544275734578/1150792660035059823/squawker.svg) good?

x9136: > are you considering the stroke as a part of the shape?

If you say 148px, that means 160-12px, where 12px is the thickness of the stroke. So 148px excluding the stroke, and 160px including the stroke. Maybe I'm confusing you. In Figma, you can check the dimensions in both cases.

> I made some changes, is [this](https://cdn.discordapp.com/attachments/842838544275734578/1150792660035059823/squawker.svg) good?

Not pixel-perfect, but within the margin of error to accept the PR. The eye is ok now.
Screenshot 2023-09-11 at 19 12 31

PR #1410, opened 2023-08-20
Add RHVoice

chipotledonut:
## Description

Add RHVoice

## Type of change

:white_check_mark:  Icon addition (non-breaking change that adds/modifies Lawnicons's icons)

### Icons added
* RHVoice (`com.github.olga_yakovleva.rhvoice.android`)


chipotledonut:
I can't pull this commit.

`Error: Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.dom4j.Attribute.getValue()" because the return value of "org.dom4j.Element.attribute(String)" is null`

Grabstertv (changes requested):
Your icon is not in line with Lawnicons guidelines. Please refer to [the guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md) and update your icon accordingly. Thanks

chipotledonut:
> Your icon is not in line with Lawnicons guidelines

Can you please tell me why not? I've seen several icons with the same font than the original. Is that the problem?

chipotledonut:
> Your icon is not in line with Lawnicons guidelines. Please refer to [the guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md) and update your icon accordingly. Thanks

I think it is done now.

Grabstertv:
> > Your icon is not in line with Lawnicons guidelines. Please refer to [the guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md) and update your icon accordingly. Thanks
> 
> I think it is done now.

Your icon is still not following the guidelines. You need to make your icon outlined. It is currently filled and is not consistent with Lawnicons' other icons which will create visual imbalance when applied.

chipotledonut:
I think I'm not following what the problem is. There are dozens of icons as this one in Lawnicons. Have you seen the RHVoice original icon?

Screenshot_20230820-171524

Are those icons outlined or filled? Do they create visual imbalance?

KTibow:
right now your icon is just a stroke around filled tinos, which makes it overly thick. using single lines (not double), you should try to match the font used in the rhvoice logo.

chipotledonut:
> stroke around filled tinos

Tinos is the font used in the original icon. I will make it single line.

chipotledonut:
Right now it is identical to the original icon. I don't know how to make it outlined without making it too big or too thick,

KTibow:
to clarify, you need to convert the path into a single line instead of closed shapes, and use stroke instead of fill. this isn't going to be easy, you may need to manually trace it or do some calculations. however as it is right now the icon has a variable width which isn't allowed.

chipotledonut:
> However, fill can be used sparingly for small details that require minimal contrast or for elements **that cannot be accurately recreated with an outline.**

KTibow:
alright, i'll see if i can create the logo using single lines then.
(also where did you get the logo from? the one on google play looks like a sans-serif font.)

chipotledonut:
Their website and fdroid.

RHVoice in the Play Store **is not the same app.**

KTibow:
i admit this could look cleaner (you could probably adjust the spacing, curves, etc pretty easily) but here it is, an outlined version
rhvoice

x9136:
> Are those icons outlined or filled? 

I did half of it. They are outlined.

The problem is that when you have a word longer than 4 letters, you have to reduce the line thickness if you're trying to keep the original look. I've come to the conclusion that if you have to make letters out of 6px thick lines, it's better not to do this shit. It won't look good. In this case it's better to look for another solution - to make the word in 2 lines or check, maybe people can shorten the name and it will be recognizable. Or look for other elements in the design of the application.

PR #1413, opened 2023-08-20
feat: Refactor PR template

SuperDragonXD:
## Description

Adds 3 templates:
* A general one for misc stuff
* An icon addition template for icon additions (with contributor checklist)
* Blank, for major changes/advanced contributions

Since GH doesn't have a UI for selecting PR templates, I've written the PR chooser text in the default template.

## Type of change

:x: Bug fix (non-breaking change which fixes an issue)
:x: Icon addition (non-breaking change that adds/modifies Lawnicons's icons)
✅ General change (non-breaking change that doesn't fit the above categories, such as copyediting)

PR #1418, opened 2023-08-21
fix checklist rendering in pr template

benteg:
## Description
Previously the checklist in the pr template used `[x] ...`

image

## Type of change

:white_check_mark:  General change (non-breaking change that doesn't fit the above categories, such as copyediting)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/icon_addition.md
+++ b/.github/PULL_REQUEST_TEMPLATE/icon_addition.md
@@ -18,6 +18,6 @@
 * App Name (linked `package.name` to `@drawable/package`)
 
 ## Contributor's checklist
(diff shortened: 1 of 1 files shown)


x9136:
Can someone tell me what effect you plan to achieve with this?

It's understandable that people will check the checkboxes without reading, but now you can rebuke them for not reading. Just why? I don't see the point of adding unnecessary actions to the dozens of actions you have to take to create and suggest an icon. I understand when everyone tries to make the process easier, not more complicated.

🙂

SuperDragonXD:
Please leave discussion regarding the checkboxes in #1413. This PR doesn't add the checkboxes, it only fixes them.

(Tbh, I don't feel like checking a box adds additional effort. It's a better reminder than a comment.)

PR #1420, opened 2023-08-22
Update 2 icons

SiMah5:
# Description

### Icons updated
* Innertune (`com.zionhuang.music/com.zionhuang.music.MainActivity`)
* addy.io (`host.stjin.anonaddy/host.stjin.anonaddy.DefaultIcon)

## Contributor's checklist
[✅] I have followed the [Lawnicons Guidelines](./github/CONTRIBUTING.md)
[✅] I have ensured that Lawnicons builds correctly
[✅] I am willing to make changes to my icons if someone suggests changes


app/assets/appfilter.xml
```diff
@@ -223,8 +223,8 @@
   <item component="ComponentInfo{com.ichi2.anki/com.ichi2.anki.DeckPicker}" drawable="ankidroid" name="AnkiDroid" />
   <item component="ComponentInfo{com.ichi2.anki/com.ichi2.anki.IntentHandler}" drawable="ankidroid" name="AnkiDroid" />
   <item component="ComponentInfo{com.ichi2.anki/com.ichi2.anki.StudyOptions}" drawable="ankidroid" name="AnkiDroid" />
-  <item component="ComponentInfo{host.stjin.anonaddy/host.stjin.anonaddy.ClassicIcon}" drawable="anonaddy" name="Anonaddy" />
```
KTibow, line 226: remove the old drawable

svgs/addyio.svg
KTibow, line 1: the stroke width isn't consistent

svgs/innertune.svg
KTibow, line 1: the top has some quality problems - reduce the gap between the two bar thingies, and maybe remove the partially transparent thing at the end

SiMah5: i redid the top part and reduced the gap between the bars, im not sure what u meant by transparent thing tho 😅 , can u show me?

svgs/addyio.svg
SiMah5: should be consistent now

svgs/innertune.svg
KTibow: sorry that was a bit confusing
here's the original icon:
image
i think as a lawnicon, it would make sense to not draw the light blue thing

SiMah5: sorry for taking so long to respond, i would prefer if we keep it as it is ( i tried it without the light blue part and i think it's better as it is ).

x9136: example

SiMah5: oh i see, okay i'll update it rn. thanks for the help

svgs/addyio.svg
x9136, line 1: Since you used a square, the icon should be 144x144px, not 160x160px. It's worth resizing or leaving just the lock. If you leave the lock, don't forget to round off the right corners to 6px.

x9136 (commented):
The addyio icon remains to be dealt with.

svgs/addyio.svg
SiMah5: i made the icon with only the lock first but i felt it's not similar enough to the original icon so i added the square. 
u23qgI6IlW
i updated the square size but if u think removing it and leaving the lock looks better i can upload the other one.

x9136: >i felt it's not similar enough to the original icon so i added the square.

There are 2 issues that are important here:
1. Will people be able to recognize from the icon what application we are talking about? It's a wide range, a full match to the original is not required.
2. What details should the icon have to match Lawnicons? Line widths, no fill, 6px rounding where it looks adequate.

It goes something like this:
Screenshot 2023-09-10 at 10 57 18
>i updated the square size but if u think removing it and leaving the lock looks better i can upload the other one.

Technically the icon fits, but the question is quality. Try to see how it will look on your phone. The problem is that too small distances between lines cuts your eyes.

SiMah5: okay i updated it again, this one looks the best on phone imo ( i really should stop guessing how they'll look based on how they look in figma 😭. ).

PR #1427, opened 2023-08-23
Check XML style

Goooler:
Closes #1417.


Goooler:
@SuperDragonXD Can we add `check-style` into CI status checks?

PR #1430, opened 2023-08-27
6 icons, quality update

x9136:
I'll make some updates to improve the icons in terms of quality and consistency with the guides.


x9136:
Damn, you've really added extra work for newbies in the process of adding icons. That sucks.

KTibow:
> Damn, you've really added extra work for newbies in the process of adding icons.

are you complaining about the PR templates?

x9136:
> are you complaining about the PR templates?

That's right. You can see from the decision that it was made by people who don't care about the convenience of the contributors. Unfortunately, in such tasks good solutions appear only if convenience is a concern.

KTibow:
while you're at it, i think it would be better if the Google Authenticator logo was single stroke instead of double stroke - that's debatable though, as the icon has some fancy overlap stuff going on

PR #1458, opened 2023-09-15
Quality update 2

x9136:


KTibow:
9136 how could you betray your standard of adding highly optimized icons
(see: transformations are present, some arcs could be removed, there are some clip paths and unused rectangles in here too)
(this comment is mostly sarcasm, you don't actually need to change anything)

x9136:
> how could you betray your standard of adding highly optimized icons

I use transforms and other things if it's faster to make an icon that way. I simplify the icon code only when exporting, with the help of a plugin. 👍

Chefski (approved):
Thanks

PR #1495, opened 2023-10-09
Quality update 5 + appfilter updates

x9136:

## Icons updated
Moon Reader Pro (`com.flyersoft.moonreaderp/com.flyersoft.moonreaderp.ActivityTxt`)
Moon Reader Pro (`com.flyersoft.moonreaderp/com.flyersoft.moonreaderp.ActivityMain`)

## Returned activities
Feedly (`com.devhd.feedly/com.devhd.feedly.Main`)


svgs/instaddr.svg
Chefski, line 1: Triangle is missing on the bottom left arrow and also the icon seems to have some kind of transparency? Its 100% but the colours are slightly transparent???
image

Chefski (changes requested):

svgs/instaddr.svg
x9136: > Triangle is missing on the bottom left arrow

Screenshot 2023-10-14 at 16 29 09
It's on purpose. You think it's better with the arrow?


> Its 100% but the colours are slightly transparent???

Screenshot 2023-10-14 at 16 33 25
In Figma, both the layer and the color have opacity.

KTibow: I would personally keep the arrow but make it so the arc is smaller so it doesn't overlap with the piece of mail

x9136: >so it doesn't overlap with the piece of mail

I tried it this way first, but there's not enough space: either the arrow would start to merge with the line and turn into a blot, or the distance between the arrow and the letter would be too small.

KTibow: I mean you would adjust the line and curve so that there's enough space (move the curve start to the left, and reduce the radius)

Chefski: Oh I just realised the pass through was at 90%. It's fine now.

And about the arrow, I think it should stay. Looks incomplete without it but maybe needs to be adjusted some way.

x9136: > you would adjust the line and curve so that there's enough space

There will be an imbalance with the right line, so it will have to be changed. This change will break the connection with the top line. The original idea is to have these lines form an overall shape. I don't want to give up this idea in my variant, because it allows to make the icon neater.

Screenshot 2023-10-14 at 17 45 41

svgs/purple_tv.svg
RaduBratan, line 1: Here's my two cents about this icon. [The original one](https://github.com/NyanArchive/OrangeTV/blob/main/app/models/src/main/res/mipmap-xxxhdpi/ic_launcher_foreground.png) looks 3D, and the Lawnicons version is also meant to evoke a similar feeling (it no longer looks 100% 3D because it's simplified). This is why I recommend the internal square should remain slightly offset to the top right. Your current implementation looks 2D-ified.

svgs/too_good_to_go.svg
RaduBratan, line 1: The new icon seems to have two endpoints that are misaligned, which create a weird spike. I checked and they are 0.001px apart on the Y axis (86.846px and 86.847px).

svgs/purple_tv.svg
x9136: Screenshot 2023-10-14 at 23 01 40
Perhaps one of these versions will be even better than your suggestion?

svgs/too_good_to_go.svg
x9136: I'm not sure what you're talking about. Could you be more specific, show a screenshot of what looks weird?

Also, I can't take into account 3 digits after the decimal point in Figma, because it rounds to two.

RaduBratan: Here's what I mean by "spike":
image

Didn't know Figma only allows 2 digits. I'll fit it myself in Illustrator, then. I noticed the same thing happened in other icons as well (such as the Play Store app).

I know it's a small difference in positions, but this is how it looks on my end:
image
image

svgs/purple_tv.svg
RaduBratan: Wow, the third one looks really nice! Even closer to the original, best version imho.

svgs/too_good_to_go.svg
x9136: @RaduBratan I see.

too_good_to_go
I manually set a slightly different value. If you don't have a spike anymore, then I'll upload that icon. If there is, then I'll delete my variant.

RaduBratan: Yep, you fixed it, now it's a single anchor point.

svgs/instaddr.svg
RaduBratan: A middle ground could be achieved by removing the external strokes of the top and right arrow heads and adding an internal stroke for the left one, but to be honest, the original version looks just fine.

Chefski: Would the arrow going along the triangle not look better? Instead of it going inwards

x9136: > Would the arrow going along the triangle not look better?

I have a feeling that either I answered above or I didn't quite understand how you imagine it. So now I'm interested in having you try it yourself and assess the situation. If you have the time, copy [a version](https://www.figma.com/file/RAXrCA2wQKRANzMGEEfJFr/Untitled?type=design&node-id=1%3A12&mode=design&t=zzxpb7Qq352Y5lQx-1) that's easier to edit.

KTibow: Something kinda like this
image
Except it goes out of bounds then

x9136: > Except it goes out of bounds then

Well, and that means the icon doesn't fit. Furthermore, the arrow turned out quite different from the others and on a small resolution on will be more like a dot.

I just wanted Chefski to try to make such an idea himself, so that he could see from his own experience if it is possible to do it properly.

x9136: Anyway, we've already spent too much time on this, so I suggest we either keep the current version — it's probably better than what's in the library. Or I'll remove my suggestion and it will be easier to move on.

RaduBratan: I have 3 things to mention about this icon:
1. on my end, in Adobe Illustrator, the icon, and every one of its components, appear to be 0.001px to wide, too tall, or too much offset to the left/right (e.g. the width of the icon isn't 148px, it's 148.001 or .002; the Y coordinate of one of the lines isn't 18px, it's 18.001, etc) (see first image)
2. the envelope is still set to 90% opacity (I remember it was mentioned before, but I don't think it has been resolved; see first and second image)
3. what if we do something like in the third image? could it serve as a middle ground to the arrows problem?
<br/>

[first image]
image
[second image]
image
[third image]
image

x9136: 1. I think it's such an inconsequential difference that without numbers, no one in the world will say there is one.
2. Pay attention to make sure that the file list includes all changes:
Screenshot 2023-10-16 at 21 29 47
3. It seems to me that if we vote, your version wins. If you're willing to offer it in your PR, I'll remove my version.

RaduBratan: Sorry for 2 then, didn't realize 😅

PR #1414, opened 2023-08-21
126 new icons, 1 update

x9136:
## Type of change
:x: Bug fix
:white_check_mark: Icon addition
:x: General change
### Updated
Moon+ Reader Pro (`com.flyersoft.moonreaderp/com.flyersoft.moonreaderp.ActivityMain`)
Moon+ Reader Pro (`com.flyersoft.moonreaderp/com.flyersoft.moonreaderp.ActivityTxt`)
### Requests
Alert (`com.ukrainealarm/com.ukrainealarm.main.MainActivity`)
AndDrop (`com.neure.anddrop/com.neure.anddrop.SplashActivity`)
Anytype Legacy (`com.anytypeio.anytype/com.anytypeio.anytype.ui.main.MainActivity`)
App Locker Master (`com.vnsolutions.fastappslocker/com.gmsolution.fastapplocker.MainActivityAlias`)
Arena Kuota (`com.arenakuota.mobile/com.w38s.StartupActivity`)
Artifact (`ai.nokto.wire/ai.nokto.wire.MainActivity`)
AstraCrypt (`com.nevidimka655.astracrypt/com.nevidimka655.astracrypt.MainActivity`)
Auto Change Wallpaper (`viet.dev.apps.autochangewallpaper/viet.dev.apps.autochangewallpaper.activities.SplashActivity`)
Autosync (`com.ttxapps.autosync/com.ttxapps.sync.app.MainActivity`)
Bar and Bench (`com.barandbench/com.zoontek.rnbootsplash.RNBootSplashActivity`)
Battery Widget (`com.lstapps.batterywidget/com.lstapps.batterywidget.MainActivity`)
BC Services Card (`ca.bc.gov.id.servicescard/ca.bc.gov.id.servicescard.screens.splash.SplashActivity`)
Bibit (`com.bibit.bibitid/com.bibit.bibitid.ui.BibitActivity`)
Bitexen Global (`com.bitexenglobal.exchangeapp/com.bitexenglobal.exchangeapp.activity.SplashScreenActivity`)
BMO Digital Banking (`com.bmoharris.digital/com.bmoharris.digital.MainActivity`)
Booksy (`net.booksy.customer/net.booksy.customer.SplashActivity`)
Botim (`im.thebot.messenger/im.thebot.messenger.BOTStartPage`)
BSRemote (`net.froemling.bsremote/net.froemling.bsremote.ScanActivity`)
[...]

PR #1512, opened 2023-10-17
Add 1 icon

SagarDevAchar:
# Description

Added 1 icon

### Icons added
* Filen (`io.filen.app`)

## Contributor's checklist


x9136:
Thanks for contributing, the icon fits.

You may want to work on the quality, so I have a few suggestions:

Screenshot 2023-10-17 at 08 05 11
- [The guides](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/CONTRIBUTING.md#foundation) recommend rounding off right angles at 6px. I suggest doing so because it allows icons to have more commonalities and look more appropriate in the library.

- Most icons use a 12px stroke as a base stroke, as described in the guides. Other thicknesses are usually used when 12px degrades the quality of the icon.

- Try to avoid small voids in the circles, because it looks like a superfluous detail in a large size, and especially on a phone screen. [An example](https://github.com/LawnchairLauncher/lawnicons/pull/1414/files#diff-92c069f5c24cddc99c9739ac25171c8a9bf4be7ca5416f83670f5fa90cbe5f8c).

P.S. If you plan to add icons in the future, I suggest you figure out how to update the commit history so that we don't have to scroll through it longer and longer.

SagarDevAchar:
> You may want to work on the quality, so I have a few suggestions:

Updated!

> P.S. If you plan to add icons in the future, I suggest you figure out how to update the commit history so that we don't have to scroll through it longer and longer.

Will fix that before my next contribution.
Apologies for the inconvenience.

svgs/filen.svg
x9136, line 1: Screenshot 2023-10-17 at 10 16 56
As you can see, if you use 12px inside, things get too dense. And that density will increase while the size decreases — on a phone screen, the size is small. On the right is an option if you leave 10px for the center part.

Also you decided to use a fill for some reason. Please use only stroke. In your case it is enough to make 10×10px circles with 10px stroke.

x9136 (changes requested):

x9136:
Perfect 👍 . There's one detail left:

Screenshot 2023-10-17 at 17 35 16

PR #1517, opened 2023-10-17
Quality update 6

x9136:


svgs/sonos.svg
RaduBratan, line 1: I wonder if the new icon could be mistaken with "SMS"?

x9136: Probably, but rarely, because if you search for SMS apps in Google Play, almost all of them have a dialog cloud icon.

RaduBratan: Fair enough.

Chefski (approved):

PR #1533, opened 2023-10-27
Optimize icons

KTibow:
SVGO + manual optimizations + multiple autooptimizations
-18kb from zipped version


Goooler (approved):

x9136 (changes requested):
Checking such PR is maximally inconvenient, because the browser not only lags when loading in the first minute, but also loads about 120 icons. The rest you have to load by yourself, one by one.

If you don't mind and will do optimization, I suggest adding no more than 150 icons to a PR. Or you should write a tool that will visually compare the icon before and after, and then tell you where there are differences. And you will show a screenshot that there are no differences in 844 / 844 icons. :D

Good job.

KTibow:
- Manual optimizations are done in an environment where I can see the changes live
  - Before commiting I can re-check each icon
- SVGO never changes icons
- My scripts check the icons are the same before making optimizations

Consider viewing commit by commit to reduce lag

x9136:
@KTibow I'm curious how exactly did the 2 changes come about?


> Consider viewing commit by commit to reduce lag

Takes a lot of extra time anyway. I checked all the icons here because I was interested in drawing conclusions in case I wanted to make a PR of several hundred new icons.

But for no reason such huge PRs will be checked without looking. Because it's laggy, because you have to move from commit to commit, because not all icons are loaded at once, and so on.

KTibow:
> I'm curious how exactly did the 2 changes come about?

Didn't see your review, I'll fix those soon, those were from when I was changing hand-drawn rounded rects to `<rect>`
> But without motivation such huge PRs will be merged without looking.

Usually I don't make errors, and they have been merged just fine in the past, especially when it's 100% automated

x9136 (approved):

PR #1563, opened 2023-11-18
Add Mój Orange

Chefski:
# Description

New icon 🤗 

### Icons added
* Mój Orange

## Contributor's checklist

### DIFF
--- a/app/assets/appfilter.xml
+++ b/app/assets/appfilter.xml
@@ -3131,6 +3131,7 @@
   <item component="ComponentInfo{uz.mobiuz.mobiservice/uz.mobiuz.mobiservice.dev.SplashActivity}" drawable="mobiuz" name="Mobiuz" />
   <item component="ComponentInfo{pl.nask.mobywatel/pl.gov.mc.fringers.mobywatel.MainActivity}" drawable="mobywatel" name="mObywatel" />
   <item component="ComponentInfo{com.axiel7.moelist/com.axiel7.moelist.uicompose.MainActivity}" drawable="moelist" name="MoeList" />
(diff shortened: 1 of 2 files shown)


svgs/moj_orange.svg
KTibow, line 1: needs full corner radius

x9136: Screenshot 2023-11-18 at 10 25 14
I'd also prefer a little more indentation between the square and the line inside to make it a little neater on the phone. Maybe it's a matter of taste.

app/assets/appfilter.xml
x9136, line 1: Perhaps it is worth changing Mój to Moj?

Chefski: I just went with what the original icon looked like.
image

Should I just make the name be the same way we have just eat, pyszne.pl etc. `Moj Orange / Mój Orange`. Autocorrect would probably change moj to mój

x9136: > I just went with what the original icon looked like.

It's a matter of style in this situation. I try to leave more space so that lines stick together on the phone screen in fewer cases. But I'll tell you about one thing, if you're interested. You made the square inside the logo shape — by doing this you "ate" the original size of the icon, but left the line inside the old size. So you actually reduced the margin that is in the logo between the edge of the icon and the line. 😄 

> Should I just make the name be the same way we have just eat, pyszne.pl etc. Moj Orange / Mój Orange

It's up to you. The option that you think will show this icon more often in search results. The question is, how will people search for the icon, "Mo..." or "Mó..."? In fact, by changing the characters to Latin we will only partially help the search. It would be better to consider o and ó, as well as other similar cases, as equivalent for the results.

Chefski: I'll make the changes soon.

I have no experience in Android development so I have no idea how to do it 😅 I'll let you do it

x9136 (approved):

PR #1577, opened 2023-11-22
Add SAQ Icon

service-paradis:
# Description

Add SAQ Icon (https://play.google.com/store/apps/details?id=com.saq.android&hl=fr&gl=US)
Inspired from the icon found in the footer of https://www.saq.com/

### Icons added
* SAQ (`com.saq.android`) 

## Contributor's checklist


svgs/saq.svg
Chefski, line 1: I think going we the icon square-ish shape could use a single line with 14px and bottom line could be 10px or 8px. I think it looks more consistent with other icons. What do you think?

Something like this:
image

Chefski (changes requested):

svgs/saq.svg
service-paradis: You're right, would probably look better. I will try to submit a proposal for it next week!

service-paradis: I updated the icon with your proposal :)

Chefski (approved):

PR #1579, opened 2023-11-23
Automatically label PRs

Chefski:
## Description

Added a workflow which will automatically label new PRs with "needs review". If the pr has changes to any .md file, the documentation label will be applied. If there's any changes to directories/files which would be changed for icons, the "icon change" label will be applied.

In summary:
**Any PR:** needs review
**`.md` files changed:** docs
**`svgs/` or `app/assets/appfilter.xml` changed:** icons
**`app/` changed:** code
**`app/src/main/res/*/strings.xml` changed:** locale

## Type of change


x9136:
> If there's any changes to directories/files which would be changed for icons, the "icon change" label will be applied.

If I understand you correctly, virtually all PRs will hang with that label? If almost all PRs have the same label, this label is almost useless. If it is necessary, it is more useful to define a label for those PRs that for some reason do not receive the "icon change" label.

Chefski:
I think specifying the icon change label helps easily differentiate the PR from another PR which doesn't apply for any other label automatically. We could add more labels but going simple for now, I think these are good. Filtering PRs which add icons will also be easier. It's mainly for organisational purposes done automatically for the most common PRs (until we add more).

Hopefully that makes sense

x9136:
> I think specifying the icon change label helps easily differentiate the PR from another PR which doesn't apply for any other label automatically.

Let's take the current situation as an example. If you're looking for PRs w/o icons, you should first see PRs that have `icon change` and then separate the others from them. If you're looking for PRs with icons, it's enough to look at `icon change`.
<details><summary>Now</summary></details>


If you're looking for PRs w/o icons, it's enough to look at `general change`. If you're looking for PRs with icons, you should first see PRs that have `general change` and then separate the others from them. 
<details><summary>Alternative</summary></details>

Since in the second case there are fewer PRs with `general change`, which in principle reflects the usual situation, the list is less overloaded with labels. Also, to find PRs with icons you almost never have to filter other PRs by label (because there are few of them), and in the current situation to find PRs w/o icons you will often have to filter many other PRs.

> We could add more labels

The most valuable information is written in the title, and if there are 2-3-4 labels next to each other, they will stand out brighter on every PR and compete with each other for attention. Perhaps the colors should be handled separately so that the labels pull less attention to themselves.

> Filtering PRs which add icons will also be easier.

True. The only thing is, if you need to enable filters, then maybe that means it's hard to use the list without them.

Anyway, these are the points I wanted to emphasize. There are also PRs that contain both icons and code changes. In the current situation it will turn out that they may be skipped, because the label will be only `icon change`.

Chefski:
Maybe we can get rid of the needs review, changes requested label to remove clutter?

> There are also PRs that contain both icons and code changes

Which is a good example of why the icon change label should be added. Without it as a reviewer, I'm to assume that it's just code change because that's the only label there.

> In the current situation it will turn out that they may be skipped, because the label will be only icon change.

All labels that apply to whatever the criteria is will get a label. If a markdown gets changed and icon is added then the PR gets both labels. We could add another label, general change for example, which will apply to anything that is in `app/` but excludes `appfilter.xml`. Sure, appfilter could have a general change like formatting. In that case the reviewer can change the label maybe.

x9136:
> Maybe we can get rid of the needs review, changes requested label to remove clutter?

I think so, but partially. You could also make the words shorter, which would improve the readability of the list. That's how it works in my head:

No label if there is a need for a review, because all requests need a review.
**сode** — if there are code changes.
**docs** — if there are *.md changes.
**icons** — if there are svgs.
**reviewed** — if the review is done and changes are needed. Shortened and renamed because it duplicates the status of the review. I wouldn't want to give it up because GitHub's writing is too inconspicuous.

So let's imagine what the PR with the most labels would look like. 
Screenshot 2023-11-23 at 22 03 35

Chefski:
> No label if there is a need for a review, because all PRs need a review. **сode** — if there are code changes. **docs** — if there are *.md changes. **icons** — if there are svgs. **reviewed** — if the review is done. Shortened and renamed because it duplicates the status of the review. I think there's a better word for it. It would be a lucky thing to choose a word that comes first or last on the list of labels. I wouldn't want to give it up because GitHub's writing is too inconspicuous.
> 
> So let's imagine what the PR with the most labels would look like. Screenshot 2023-11-23 at 22 03 35

Yeah I think those are good changes. So we'll shorten the names?

x9136:
> So we'll shorten the names?

Go ahead.

The only question I still have, but I haven't come up with a solution yet, is when there is the `reviewed` label and 2-3 others. In such a case it's not clear what the `reviewed` refers to — icons, code, docs or everything. That is, the task is somehow to say what's left to review.

We could agree to remove the label on which the review was made. For example, there are `icons` and `code`. You have reviewed the icons, removed the `icons` label, and the `code` remains unreviewed. This flow feels too different from the case with one type of label. And it can be confusing because statuses don't have a partial review label.

Adding different labels (`icons` and `icons ☑`, etc) or making a checklist for review is also a poor option because it requires additional management.

So, until there is a better solution, I suggest doing nothing about it and assuming that the case is rare. But I wouldn't set the `reviewed `label on a PR if only a partial review is done. Most likely the solution here will not be label-related: make a recommendation to split a PR with icons and code into two, call a reviewer separately when you've only done your part, and so on. 

How would you show what's left to review in a multi-label PR?

Chefski:
I think you are thinking too hard on this. Looking at the previous labels, I think `reviewed` should be left out as it implies the code has been fully reviewed and there's no changes needed. `changes requested` can be left on top of `needs review`. We can request reviews from other reviewers if needed.

Generally, icon and code change PRs are separate. As a small repo which mainly gets only icon changes, I think this works fine.

x9136:
> `changes requested` can be left on top of `needs review`.

I agree about `changes requested`, but I don't see any practical benefit of `needs review`. A PR either needs review or needs changes, so suffice to leave the second one. Either I misunderstood you, or you see some benefit from `needs review`.

Chefski:
Don't think there's much benefit to it either.

> That is, the task is somehow to say what's left to review.

It only slightly fixes the problem of knowing what's not been fully reviewed.

x9136:
In summary: `code`, `docs`, `icons`, `changes requested`. Maybe a little less saturated colors. I have nothing to add.

Chefski:
> Is there a better name for changes requested so as not to duplicate the GitHub status? I would suggest wip. Or do you want to remove it altogether?

I guess `wip`, `rework` or `cr` would be some options. I think keeping it will help with keeping track of progress and somewhat indicate that the PR has been reviewed by someone.

x9136:
> I guess `wip`, `rework` or `cr` would be some options. I think keeping it will help with keeping track of progress and somewhat indicate that the PR has been reviewed by someone.

To me `wip` and `rework` are equally good because they speak for themselves.

Chefski:
Although `wip` is nice and short, I think `rework` would be better suited here as it's more clear that something needs change. `wip` gives more of a draft PR impression as it stands for "work in progress".

KTibow:
you could literally convert the pr to a draft also

x9136:
It would also be handy if `rework` set itself when you finished a review with a `Changes requested` status.

Chefski:
Yeah tried making an action but can't get it to work right. Couldn't find any marketplace action that would also do it.

Chefski:
Gonna merge this and we can discuss the changes for `general_changes.md` on Discord

PR #1609, opened 2023-11-30
Update search message and add Marathi translations

dhirenvasnani:
## Description
- Added Marathi translations
- Changed Search message to "**Search through %d icons**" from "**Search %d icons**" for clarity

## Type of change


x9136 (approved):
I don't know Marahti, but the translator didn't show anything criminal. As for the search hint, both options are quite clear and close in meaning, so either option is fine.

PR #1611, opened 2023-12-01
Make "External Links" a translatable String

Chefski:
## Description
Made "External Links" a translatable String and added the Polish Translation.

"Icon Request Form" remains as the last String which I couldn't get to work.

## Type of change


service-paradis:
Great, it would be "Liens externes" in french :)

Chefski:
@service-paradis thanks! Added it :)

SuperDragonXD (approved):
I'll migrate the other strings to `strings.xml` later, but looks good.

x9136:
Just in case, I will draw your attention to the idea that it is better to keep all translations in one file, because it will be easier to add, edit or delete lines.

PR #1627, opened 2023-12-05
Create Pull Request

### LINKED ISSUE #1626: [DISCUSSION] Workflow to optimize SVGs is not working
### About this discussion

If you take a look at https://github.com/LawnchairLauncher/lawnicons/actions/workflows/optimize_svgs.yml, every workflow fails.

It should be fixed or removed if not necessary.

validcube:
## Description

Create a Pull Request instead of pushing directly to repository (See relevant rule: GH006).

Fixes #1626 

## Type of change

✅ Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,15 @@
+name: Auto-merger
+
+on:
+  pull_request:
+
+jobs:
+  optimize-auto-merge:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.title == 'Optimize SVG(s)'
+    steps:
+      - env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+        run: |
+          gh pr comment $PR_URL --body "No issues found, auto-merging the pull request."
+          gh pr merge $PR_URL --auto --squash --delete-branch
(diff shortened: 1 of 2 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,15 @@
+name: Auto-merger
+
+on:
+  pull_request:
+
+jobs:
+  optimize-auto-merge:
```
validcube, line 7:
```suggestion
  merge-optimize-svg:
```

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.title == 'Optimize SVG(s)'
```
validcube, line 9: Checking only for the title is not ideal

See https://github.com/LawnchairLauncher/lawnchair/blob/13-dev/.github/workflows/ci.yml#L109-L119 for a better implementation

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,15 @@
...
+    steps:
+      - env:
+          PR_URL: ${{ github.event.pull_request.html_url }}
+        run: |
+          gh pr comment $PR_URL --body "No issues found, auto-merging the pull request."
+          gh pr merge $PR_URL --auto --squash --delete-branch
```
validcube, line 15: Might be a problem, because we would need to let it run a bit for the required check to complete then we can merge it.

validcube (commented):
bleep: self-review

PR #1636, opened 2023-12-06
Add request_form french translation

service-paradis:
## Description

Add french translation for `request_form`

## Type of change

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -8,7 +8,7 @@
 
     <!-- The hint shown in the search bar, with %d as the number of icons. -->
     <!-- We don't currently provide a singular form, because it's always guaranteed that there will be atleast 2 or more icons. -->
-    <string name="search_bar_hint">Rechercher %d icônes</string>
+    <string name="search_bar_hint">Rechercher parmi %d icônes</string>
 
     <!-- Text shown when there are no search results-->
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-fr/strings.xml
```diff
@@ -18,11 +18,15 @@
...
     <!-- Popup menu item that links to the About page -->
     <string name="about">À propos</string>
-
-    <!-- About page -->
     <!-- External Links label -->
     <string name="external_links">Liens externes</string>
 
+    <!-- External Links labels -->
+    <!-- Icon Request form link -->
+    <string name="request_form">Formulaire de demande d\'icône</string>
```
Chefski, line 26: Out of curiosity, is the `\` supposed to be in it or is it a misstype?

service-paradis: I'm pretty sure we need to escape single quotes. I did the same things for other strings before:

```xml
    <!-- Subtitle displayed below the user who inspired the app icon -->
    <string name="special_thanks_icon">Pour avoir inspiré l\'icône de l\'application.</string>
    <!-- Subtitle displayed below the user who named the app -->
    <string name="special_thanks_name">Pour avoir donné le nom à l\'application.</string>
```

Chefski: Completely forgot how common ' is in French

PR #1642, opened 2023-12-07
Add Mobvoi

validcube:
# Description

Add Mobvoi, I have no idea what I'm doing please help.

### Icons added
* Mobvoi (` com.mobvoi.companion.aw/com.mobvoi.companion.aw.StartEntry`)

## Contributor's checklist


svgs/mobvoi.svg
x9136, line 1: Screenshot 2023-12-10 at 12 29 27
Incorrect size, it's worth enlarging the icon so that the height becomes correct.

In general, 8px stroke is good for very dense icons. In your case, the icon has very little detail and lines, so if you view it on your phone screen, it will stand out because of the 8px. Formally you did it right, but if you have the desire, it's easy to do better.

x9136 (changes requested):

svgs/mobvoi.svg
validcube: Should be correct now

image

validcube: > In general, 8px stroke is good for very dense icons. In your case, the icon has very little detail and lines, so if you view it on your phone screen, it will stand out because of the 8px. Formally you did it right, but if you have the desire, it's easy to do better.

I could do something like this, lmk what you think:

image

x9136: Screenshot 2024-01-07 at 14 15 17
It feels like you moved the icon down before saving.

<details><summary>Open a screenshot</summary></details>
As you can see, the icon is noticeably out of style due to the fact that it's created with 8px stroke.

x9136: > I could do something like this, lmk what you think:

The most important thing is to use 12px stroke. Then any of these options will look good. I already showed my option in Discord a month ago.

x9136 (approved):

PR #1656, opened 2023-12-10
Add i2pd

Sir-Photch:
# Description

https://github.com/PurpleI2P/i2pd-android

### Icons added
* App Name (`org.purplei2p.i2pd`)

## Contributor's checklist


Chefski (commented):
I'm not sure if the icon is not a little too thin. Is there any other icon ideas that would be possible for this app? Did you check how it looks on the phone? If you could show a screenshot it'd be nice

Sir-Photch:
Hm, I think you have a valid point there. I guess I could do "mask only". Would be a little generic then though don't you think? Let me know your thoughts. I will try to come up with something else in the meantime.

x9136:
try

Sir-Photch:
looks interesting, but I am not sure if variable length lines fit the contributing guidelines?

x9136:
> but I am not sure if variable length lines fit the contributing guidelines?


Screenshot 2023-12-13 at 20 37 41
Technically, everything fits the guides.

Sir-Photch:
I tried to be a little creative; I didn't want to give up the "structure" around the mask. What do you think? I admit this may not necessarily look like other lawnicons x) (but it does fit the guides, amirite?)

x9136:
> What do you think?

The latter version and the one with just the mask are too different from the original. The reason is that the icon has a very expressive jester's cap.

Screenshot 2023-12-22 at 20 17 51
Stopping at this option, there are 2 comments to it. 1. Actually you made a square icon, so it should be smaller according to the guides. 2. In terms of quality, there are a lot of small spacing between strokes — it looks sloppy on the phone screen.

Sir-Photch:
here goes another try

x9136:
> here goes another try

Screenshot 2024-01-03 at 23 42 28
The only thing that remains is the wrong stroke thickness. I suggest using 10px.

Sir-Photch:
right. I corrected that.

x9136 (approved):

PR #1668, opened 2023-12-14
Updating and simplifying templates

x9136:
## Description
I suggest some changes to the templates to remove unnecessary information and clarify what is needed from the contributor.

## Type of change

## Changelog
- Removed [DISCUSSION] from the issue template because it was duplicating the label.
- Shortened the subheadings in the icon addition template because they were duplicating the header.
- Removed the note about the icon being added from the request form, because that wasn't used and you can see in the current requests table what requests are fulfilled.
- Removed the indentation in the icon lists to make it easier to read.
- Renamed the package name example to make it closer to the real ones.
- Simplified the entry to add links because the `@drawable/package` part was unclear.
- Clarified in the PR template what actions to take. In particular it was not obvious that the template could be inserted into the description if you clicked on the link.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/icon_addition.md
+++ b/.github/PULL_REQUEST_TEMPLATE/icon_addition.md
@@ -1,21 +1,20 @@
 # Description
-<!-- Please provide a short summary of what icons you added, changed, or linked
--->
+<!-- Please provide a short summary of what icons you added, changed, or linked -->
 
 ## Icons addition information
-<!-- Please specify if you added an icon that was requested in the icon request form, as seen below -->
-### Icons added
-* App Name (`package.name`) - added from icon request form
-* App Name (`package.name`)
-* App Name (`package.name`)
 
-### Icons updated
-* App Name (`package.name`)
-* App Name (`package.name`)
+### Added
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
x9136, line 1: Screenshot 2023-12-14 at 22 50 19
I'm not sure it works and is needed, so I suggest deleting it.

.github/PULL_REQUEST_TEMPLATE/icon_addition.md
x9136, line 1: The preview doesn't take line breaks into account, but I think it should be fine since the [General change template ](https://github.com/LawnchairLauncher/lawnicons/blob/develop/.github/PULL_REQUEST_TEMPLATE/general_change.md) works fine once selected.

.github/pull_request_template.md
SuperDragonXD: It won't work anymore because the `icon addition` label doesn't exist. Alongside that, it may need `triage` or higher permissions to add those labels.

SuperDragonXD (approved):

Chefski (approved):

PR #1673, opened 2023-12-15
Add icon for Benny & Co.

service-paradis:
# Description

Add icon for Benny & Co. (https://play.google.com/store/apps/details?id=com.bennyco.app&hl=fr_CA&gl=US)

### Icons added
* Benny & Co. (`com.bennyco.app`)

## Contributor's checklist


app/assets/appfilter.xml
x9136, line 1: A better name would be "Benny and Co.".

svgs/benny_co.svg
x9136, line 1: Screenshot 2023-12-17 at 07 10 13
The whole rooster is best made less detailed, but really the most important part is the comb. So I would try to maximize it to make it really expressive.

x9136 (changes requested):

app/assets/appfilter.xml
service-paradis: Since it was founded by French Canadians, I'm not sure about replacing the ampersand with "and". It still mostly used in Québec where people mostly speaks french. They recently opened a few restaurants in Ontario (where people speaks mostly english).

The ampersand is a great way to be understood by people speaking both french or english.

x9136: My premise is that it's more habitual to write "Benny and" in a search query than to recall how to type an ampersand. 

If you think that most people are used to typing the ampersand in this case because the app name is heavily tied to it, then no problem.

x9136:
Most likely, this PR will have to be recreated. If you do this, you can use this icon.
benny_co

PR #1727, opened 2024-01-04
Add icon for Flashfood

service-paradis:
# Description

Add icon for Flashfood (https://play.google.com/store/apps/details?id=com.flashfoodapp.android&shortlink=73747a82&c=downloadpage_playstore&pid=website&source_caller=unknown)

I took the icon from the footer at https://www.flashfood.com/

### Added
Flashfood (`com.flashfoodapp.android`)

## Contributor's checklist


x9136:
Looks legit.

Nevertheless, since there's a challenging curlicue in the icon, it's always a candidate for simplification. What do you think of the next option?

Screenshot 2024-01-04 at 19 33 56

service-paradis:
Good idea! Is it better?

x9136:
> Good idea! Is it better?

Overall yes, although it's cramped.

I'd like to draw your attention to one thing. If you make a stroke inside the shape of the logo (as inside the ellipse here), it turns out that you cut off part of the ellipse with the stroke. But at the same time, you made the leaf size even larger than in the original. As a result, the leaf is cramped inside the circle.

The icon becomes sloppy, and the quality decreases because of the small distances between the strokes. It is worth considering these things.

Screenshot 2024-01-04 at 22 55 39

service-paradis:
Thanks for the feedback, you're right, it was not coherent to use a standard centered outline for the leaf and an inner outline for the circle around.
I think this new version should be better!

PR #1752, opened 2024-01-14
linked 18

SKAZE01:
## Click on the template that fits your PR
* [**Icon addition**](?expand=1&template=icon_addition.md)
* [General or miscellaneous change](?expand=1&template=general_change.md)
* *[Blank](?expand=1&body=+) (not recommended)*


x9136:
Hey!

To add a link, you must already have an application icon in Lawnicons. You shouldn't add links to icons that seem to fit, because those icons belong to completely different applications. This will cause confusion.

In the case of My Android Tools, you just replaced the links with your own, rather than adding additional ones. Also, the icons of default and pro versions are just as different.

If you want to contribute, I suggest spending more time on how things work: look at other pull requests, read Lawnicons guidelines.

PR #1758, opened 2024-01-16
Quality update 12

x9136:
Making icons closer to Lawnicons style.


x9136:
This will be the last quality update from me for the foreseeable future, because the whole series of such updates is mostly needed to clean up moments that have passed the indulgent reviews. This is a labor that should be avoided by carefully checking every PR.

I encourage reviewers to be stricter in future reviews, otherwise Lawnicons quality will drop.

PR #1755, opened 2024-01-15
Lawnicons Guidelines v2.2

x9136:
Adjusted the requirements for square icons, gave samples and clarified wording with @Grabstertv.

## Type of change
:x: Bug fix (non-breaking change which fixes an issue)
:white_check_mark: General change (non-breaking change that doesn't fit the above categories, such as copyediting)

v2 2 Changelog
[Lawnicons Guidelines in Figma](https://www.figma.com/community/file/1227718471680779613/lawnicons-guidelines)

PR #1763, opened 2024-01-19
Square icons v2.2

x9136:
Updating the square icons after #1755.

Since I was checking many of the original icons, some icons had to be updated because the original was different.


x9136:
Before merging this request, #1765 needs to be merged. If the requirements change, I'll make changes here as well.

Almost all the square icons have been redone. I'll change a couple more later as part of **Quality update 12**.

Chefski (approved):

PR #1765, opened 2024-01-19
Lawnicons guidelines v2.2.1: Simplify icon size selection and increase the variety of square icons

x9136:
**Icon size selection** 
In the previous version, I didn't take into account that there are icons that can be considered both square and non-square. In the new version, I added a brief instruction so that contributors will always know what size to choose for an icon. It will also make the review process a bit easier, since it is easier to estimate the density of an icon than the squareness.

**Rounded corners**
It's very boring if all right angles are rounded by 6px because it increases monotony by making most square icons the same. That's why I suggest rounding the corners by at least 6px, and the maximum should be not more than 32px and depend on how rounded the corners of the original icon are (at >32px, the icon is already starting to look smaller).

## Type of change


Chefski (approved):

x9136:
If everything remains unchanged until February 5, I will merge the PR, and possible clarifications will be added in other PRs.

PR #1778, opened 2024-01-24
Add 3 icons

Hamster45105:
# Description

Added 3 icons (Qantas, floccus, what3words)

### Added

Qantas (`au.com.qantas.qantas`)
floccus (`org.handmadeideas.floccus`)
what3words (`com.what3words.android`)

## Contributor's checklist


svgs/floccus.svg
x9136, line 1: Screenshot 2024-01-24 at 17 22 55
In cases when the thickness of the initial shape is close to the base stroke thickness, it is better to do with a single line instead of outlining the shape. This way it is easier to make an icon and it will look neater.


Also, try to avoid too small distances between lines, because they stick to the phone screen.

svgs/qantas.svg
x9136, line 1: The icon is thin. If you use a stroke thickness of 8px, it will stand out a lot among other icons. It's worth increasing the stroke width to 10px (don't forget to correct the size afterward).

svgs/what3words.svg
x9136, line 1: 12.06px stroke doesn't fit. Also, you made double lines, so actually your stroke became ~24px.

Screenshot 2024-01-24 at 17 39 47
14px stroke would be fine for this icon.

x9136 (changes requested):
Hey! I've left a few comments.

By the way, [Lawnicons guideline](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#contributing-icons) was updated recently — if you have time, take a look. The changes mainly concern square icons, which are not in this PR.

Hamster45105:
The icons should be fixed now!

svgs/what3words.svg
x9136: 12px stroke is not enough for this icon, as there are few shapes in it.

If interested, read about the reasons in [the Maintaining Visual Balance](https://github.com/LawnchairLauncher/lawnicons/blob/develop/docs/images/creating-icons-4-maintaining-visual-balance.png).

svgs/floccus.svg
x9136, line 1: For some reason you decided to use a fill instead of a stroke, that option won't fit. Also, we don't make icons that have lines with square ends. Even if you find such icons, they should be redesigned.

[About stroke and rounded ends](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#foundation)

x9136 (changes requested):

svgs/floccus.svg
Hamster45105: Sorry about that! Both icons should be fixed now and ready to go.

Hamster45105:
I fixed both icons, so they should be ready to go now!

svgs/what3words.svg
Hamster45105: Sorry about that, I forgot to increase the stroke width. It should be all good now.

x9136 (approved):
Floccus icon turned out to be unbalanced because of the difference in stroke thicknesses of the cloud and tick, as well as because of too small distances between them. But it can be accepted.

I'll wait until tomorrow before merging the PR in case you want to make changes.

Hamster45105:
Thanks, I'm happy, for everything to be merged now.

PR #1790, opened 2024-01-28
TL;DR for icon creation

x9136:
Since it takes a high level of interest to read the Lawnicons guidelines, I suggest making it a little easier for those who don't like to read. In theory, this will allow new contributors to make fewer trivial mistakes, as well as simplify the reviews.

## Type of change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,11 @@ For additional information on designing icons and samples, see [the Lawnicons Fi
 
 Need help? [Join Lawnchair on Discord](https://discord.com/invite/3x8qNWxgGZ).
 
+### TL;DR
+The canvas is `192×192px`. The content area for most icons are `160×160px`, but `154×154px` for the squares. No fill, the stroke width is `12px`. The shapes should be black `#000` with rounded ends and joins. Round 90° angles by `6-32px`. Maintain visual balance. Simplify details, but don't lose recognizability. Provide original and localized names, so the icons can be found.
+
+It's better to make few high-quality icons than many crappy ones and waste the reviewer's time on comments and your own on fixes.
+
 ### Canvas & Sizes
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -10,6 +10,11 @@ For additional information on designing icons and samples, see [the Lawnicons Fi
 
 Need help? [Join Lawnchair on Discord](https://discord.com/invite/3x8qNWxgGZ).
 
+### TL;DR
+The canvas is `192×192px`. The content area for most icons are `160×160px`, but `154×154px` for the squares. No fill, the stroke width is `12px`. All shapes should be black `#000` with rounded ends and joins. Round 90° angles by `6-32px`. Maintain visual balance. Simplify details, but don't lose recognizability. Provide original and localized names, so the icons can be found.
+
+It's better to make few high-quality icons than many crappy ones to save the reviewer's time and reduce possible rework.
```
Chefski, line 16: Personally I think this line is kind of irrelevant

x9136: > Personally I think this line is kind of irrelevant

Why?

The wording could be different, but I want to increase the chance that contributors won't start making icons that don't fit at all. It's too easy to waste time on it, and the PR still won't be accepted or there will be a bunch of comments that almost nobody finalizes in such cases.

Imagine a contributor made 50 inappropriate icons — wasted their time, wasted the reviewer's time checking and commenting. Instead, he could have saved time doing nothing or made 5-10 appropriate icons.

Or, let's say the contributor was focused on quantity and ends up with 15 out of 50 icons not fitting. Then the reviewer needs to check each case and explain the reason. Although the contributor could only send 35 suitable icons or make 5 more.

x9136: Anyway, if you start to think it's important, you'll add the text later.

PR #1808, opened 2024-02-02
Add icon for Pokémon TCG Live

service-paradis:
# Description

Add icon for Pokémon TCG Live (https://play.google.com/store/apps/details?id=com.pokemon.pokemontcgl&hl=en_US)

To keep things simple, I only kept a single card from the original icon (the right one). I think it is recognizable enough while keeping it minimal like the other icons in this package.

### Added
Pokémon TCG Live (`com.pokemon.pokemontcgl`)

## Contributor's checklist


x9136:
There is an unforced imbalance due to the difference in the stroke width of the card and the poke ball. To fix it, just widen the card a little and apply the 10px stroke to the poke ball.
Screenshot 2024-02-11 at 12 34 54

service-paradis:
> There is an unforced imbalance due to the difference in the stroke width of the card and the poke ball. To fix it, just widen the card a little and apply the 10px stroke to the poke ball.

Something like this?

x9136:
> Something like this?

Yes, except the stroke width is wrong.
Screenshot 2024-02-12 at 16 53 57

service-paradis:
Yes, sorry, I saw it just after I commented! Should be better now!

PR #1871, opened 2024-03-03
Added +2 icons and +2 links

sujoybyte:
# Description

### Added
Mini Mini Farm (`jp.coffeebreakin.app.minifarm`)
Haak (`com.blingame.haaktestb`)

### Linked
Mini Mini Farm (`jp.coffeebreakin.app.minifarm` → `mini_mini_farm.svg`)
Haak (`com.blingame.haaktestb` → `haak.svg`)

## Contributor's checklist


sujoybyte:
Note: Haak's icon is inspired by it's earlier logo design as it was less detailed

svgs/haak.svg
x9136, line 1: Screenshot 2024-03-08 at 09 54 28

I suggest working on the eyes to make the icon balanced. In the current version, the imbalance is caused by the difference between the stroke widths and the unnecessarily small distance between the shapes.

When you make a stroke inside the shape of the original image (for example, inside the face like here), the stroke width "eats up" the size of the shape (the face), so it's worth compensating for this, rather than making the contents of the shape as in the original.

svgs/mini_mini_farm.svg
x9136, line 1: Screenshot 2024-03-08 at 10 13 06
The 6px stroke is only used for small details. Otherwise, it is worth working on the balance by choosing closer stroke widths and increasing the distances between the shapes.

x9136 (changes requested):

svgs/haak.svg
x9136: > I thought filled shapes will not be accepted.

You're right. It's all done with a stroke.

sujoybyte: @x9136 have a look if its proper now.

x9136: It is better not to use a fill when almost the same thing can be done with a stroke.

Screenshot 2024-03-09 at 16 46 02

sujoybyte: I did it exactly like you said.

PR #1879, opened 2024-03-07
Icon proposal for LEGO® Builder

service-paradis:
# Description

Icon proposal for LEGO® Builder (https://play.google.com/store/apps/details?id=com.lego.legobuildinginstructions)

### Added
LEGO® Builder (`com.lego.legobuildinginstructions`)

## Contributor's checklist


service-paradis:
This s the version submitted in this PR:
image

This is another option:
image

Which one would be better? Or is there another option that would look best?

x9136:
Screenshot 2024-03-08 at 11 12 09

I would simplify the connections even more to avoid blackness. If this option would not work, I would reduce the number of shapes to 2 or 1, but make the connections as similar as possible to LEGO.

x9136:
> Maybe I could keep only those 2 parts with connections similar to y 2nd proposal?

It's up to you. Given the limitations of the style, it's probably best to make one lego piece that leaves no doubt that it's LEGO.

service-paradis:
Thank you for your comments. I just sent a new simplified version with only 2 shapes and reduced corner radius. I also have the following version on hand:
image

x9136:
Blackness is worth avoiding. Don't forget to round the bottom corners.

Screenshot 2024-03-11 at 18 44 39

service-paradis:
Sure, just pushed rounded bottom corners!

PR #1898, opened 2024-03-16
Add OpenStop

Sir-Photch:
# Description
https://github.com/OPENER-next/OpenStop

### Added
App Name (`de.tu_chemnitz.etit.sse.openstop`)

## Contributor's checklist


svgs/openstop.svg
x9136, line 1: In this case, it is better to use 12px stroke either for the entire icon or only for "H", because the density at the top is too high.

x9136 (changes requested):

x9136 (approved):

PR #1912, opened 2024-03-22
Add 4, Link 1

Hamster45105:
# Description

Added 4 new icons and linked 1

### Added
Install with Options (`com.zwander.installwithoptions`)
AutoWeb (`com.joaomgcd.autoweb`)
Flowit! (`com.bytehamster.flowitgame`)
Kobo Books (`com.kobobooks.android`)

### Linked
aShell You (`in.hridayan.ashell` → `ladb.svg`)

## Contributor's checklist


svgs/flowit.svg
x9136, line 1: It's a bit far from the original, although if you look at the game itself, the icon fits. I suggest 2 options:
1. Leave it as it is, but you need to raise the arrow a little so that it is in the center, not at the bottom.
2. Make the version closer to the original.

Screenshot 2024-03-22 at 16 31 19

svgs/install_with_options.svg
x9136, line 1: The size is too small, especially for a circle. I suggest trying to make a perfect circle and increase the distance between the lines, slightly simplifying the shape.
Screenshot 2024-03-22 at 16 36 45

svgs/kobo_books.svg
x9136, line 1: Although you forgot to round the right corners, I would suggest simplifying everything and using a 14px stroke.

The width of the lines in the original is too close to the base stroke width, so it is better to use 3 lines for the "K". Also, because of the 10px stroke, the icon would stand out from the rest.

Screenshot 2024-03-22 at 16 52 55

x9136 (changes requested):

svgs/flowit.svg
x9136: 1. 13px stroke width isn't allowed.
2. Corner radius for 90° angles is 6—32px.
3. It's not worth using 8px stroke in such an icon, because the difference between 12px and 8px breaks the balance, and you can make the icon look neat by limiting the strokes to 12px and 10px.

[The Lawnicons guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#tldr)

Hamster45105: > Corner radius for 90° angles is 6—32px.

I've made it 6px, are there any specific requirements on this, or just whatever looks good?

Hamster45105: > Corner radius for 90° angles is 6—32px.

Yeah sorry I forgot about that.

I've made it 6px, are there any specific requirements on this, or just whatever looks good?

x9136: > I've made it 6px, are there any specific requirements on this, or just whatever looks good?

I do it depending on the original. If the original is not rounded at all, then 6px, and if the rounding is larger, then I do it as close as possible to the original.

PR #1931, opened 2024-03-29
Add 6, Link 1

Hamster45105:
# Description

### Added
Samsung Studio (`com.sec.android.app.vepreload`)
SqueezePlayer (`de.bluegaspode.squeezeplayer`)
Microsoft Translator (`com.microsoft.translator`)
SunSmart (`au.org.cancervic.globaluv`)
Nixplay (`com.creedon.Nixplay`)
PrinterOn (`com.printeron.droid.phone`) + (`com.printeron.android.printplugin`)

### Linked
Link to Windows (`com.microsoft.appmanager`) --> microsoft_link

## Contributor's checklist


svgs/samsung_studio.svg
x9136, line 1: Screenshot 2024-03-29 at 12 00 15
It is worth rounding the corners of the pencil, taking into account the Lawnicons guidelines, and then adjust the size of the icon.

svgs/squeezeplayer.svg
x9136, line 1: The fill should be removed. The circles are too tiny, it won't work that way in the Lawnicons guidelines. The margins between them and the large circles are also too small. Also, the arrow is not centered.

Screenshot 2024-03-29 at 12 17 07
If you reduce the stroke width of large circles to 10px, align the arrow and correct the size of small circles, you will get something like this.

Screenshot 2024-03-29 at 12 15 09
If you leave 12px stroke width only at the outer circle, align the arrow and correct the size of the small circles, then it will turn out something like this.

svgs/sunsmart.svg
x9136, line 1: It is worth using 14px stroke width, because the icon has very few elements.

x9136 (changes requested):

app/assets/appfilter.xml
x9136, line 1: > com.printeron.android.printplugin

name="PrinterOn" → name="PrinterOn Print Service"

svgs/squeezeplayer.svg
Hamster45105: I have fixed the circle sizes and centered the arrow.

Removing the fill results in a donut like shape, so is it OK to leave the fill in this case?

Hamster45105:
Everything is fixed now.

It appears some of the stroke width/shape sizes problems had to do with me optimising the SVGs. Before they were optimised, everything was fine. However after I optimised them some stroke widths and sizes of some of the shapes changed.

svgs/microsoft_translator.svg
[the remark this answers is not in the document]
x9136: Screenshot 2024-03-30 at 07 26 54

Right now, the icon is pressed to the left edge (visually), and the distance between the Chinese character elements is too small. I suggest at least reducing the stroke width of the Chinese character to 10px.

It may take a lot of time to carefully align such abstract shapes, so if something else can be corrected, it's good, but another time it will be better to redesign this icon, taking as a basis only the meaning and position of elements, not the original shapes.

svgs/squeezeplayer.svg
x9136: After reducing the width of the stroke, you forgot to increase the main circle, as a result the icon became too small.

Screenshot 2024-03-30 at 07 41 03
As for the small circles, in the examples I suggested, they are all made as 6×6px ellipses with 6px stroke width. Your options for small circles are not yet in line with Lawnicons guideline.

svgs/microsoft_translator.svg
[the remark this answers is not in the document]
Hamster45105: Looking better now

svgs/squeezeplayer.svg
x9136: > The circles are too tiny, it won't work that way in the Lawnicons guidelines. The margins between them and the large circles are also too small.

Screenshot 2024-03-30 at 09 00 11
If you make the distances between the elements too small, then the elements will stick together on the phone screen, because the icons there are usually much smaller than on the computer screen. That's why I initially drew attention to this issue.

x9136 (approved):

x9136:
Cool, thanks.

Hamster45105:
No problem, thanks for the help!

PR #1948, opened 2024-04-01
+2 updates, Use official Mihon logo

AntsyLich:
# Description
First of all I would like to mention that I'm the maintainer of Mihon. 

I just recently updated lawnicons and noticed that a logo for Mihon was added. The one added was quite different from the official adaptive icon we have. 

I've read through the guideline but not sure if the icon is following it. Would appreciate feedback and I can ask the icon designer to make changes.

### Updated
Mihon (`com.mihon`)
Mihon (`com.mihon.debug`)

## Contributor's checklist


x9136:
Hey!

> The one added was quite different from the official adaptive icon we have.

To make a Lawnicons-style icon, you need to think about how to use the Lawnicons guidelines and keep the icon recognizable, not how to make the icon as close to the original as possible without breaking the guidelines. As a result, we have the current version, because it is simpler and neater.

I can explain the disadvantages of your option, they mainly relate to maintaining visual balance, but the main question is different. Do you think that after seeing the current icon without the name of the app, Mihon users will not know which app the icon belongs to? In my opinion, everyone will recognize the icon, but if you disagree, I would like to hear the reasons. 

The icons differ only in that one has a hieroglyph made of lines, and on the other the same hieroglyph is outlined. It generally doesn't affect recognition.

AntsyLich:
> Hey!

Hi 👋

> I can explain the disadvantages of your option, they mainly relate to maintaining visual balance, but the main question is different. Do you think that after seeing the current icon without the name of the app, Mihon users will not know which app the icon belongs to? In my opinion, everyone will recognize the icon, but if you disagree, I would like to hear the reasons.

Fully agree with you on this. What I'm looking at is giving a similar icon for people using native themed icon or lawnicons. Even Mihon's predecessor Tachiyomi (eu.kanade.tachiyomi) and it's other forks has a outlined icon instead of simple lines. Besides that I personally like the outlined one better 😅.

x9136:
> Besides that I personally like the outlined one better 😅.

Screenshot 2024-04-01 at 13 31 55
Here I explain the reasons why option 1 was originally made. I would leave option 1 because it will look good in any size, but I suggest voting for users. If there are 10 more votes for the replacement, then ok. For example: 👍 20 • 👎10. I'll be back on Friday and either finish this PR or close it.

**Should we replace 1 with 3?**

x9136:
The result is obvious, thanks for contributing.

kulmegil:
I mean, if the 1) was as pleasantly thick and round as 3) outline suggests it would be more of a competition...

x9136:
To do this, you need to change the Lawnicons guidelines. If you change the guidelines, then some of the icons will have to be redone and, most likely, this will lead to a decrease in the overall consistency of Lawnicons.

PR #1972, opened 2024-04-06
Add 5 Icons

Hamster45105:
# Description

Added 5 icons

### Added
Bonza Word Puzzle (`air.au.com.minimega.bonza`)*
Bin Reminder (`com.retfunsoft.binreminder`)
DAIKIN Mobile Controller (`ao.daikin.remoapp`)
Sony | Auto Play (`jp.co.sony.hes.knock`)
LiveATC (`net.liveatc.liveatc_app`)

\* 90° angles not curved because it spoils the shape. Additionally, an alternative is to have this without the dots, they symbolise the letters that are in the original icon.

## Contributor's checklist


svgs/bonza_word_puzzle.svg
x9136, line 1: Screenshot 2024-04-07 at 01 36 45
It is better to connect these shapes to avoid artifacts at the intersection and an unnecessarily thick line.

svgs/daikin_mobile_controller.svg
x9136, line 1: Screenshot 2024-04-07 at 01 34 55
It is worth connecting such points so that the lines do not fall out of each other.

x9136 (changes requested):

svgs/sony_auto_play.svg
x9136, line 1: Screenshot 2024-04-07 at 11 05 45
The height is too small, and the legs are of different widths for some reason. I don't know why the circle is squared on the sides, and the legs are rounded, but I suggest doing the opposite — as in the original.

x9136 (changes requested):

x9136:
It is surprising that pixel-perfect is difficult in Inkscape. Almost all the icons that were made there have fractional numbers.

Hamster45105:
> It is surprising that pixel-perfect is difficult in Inkscape. Almost all the icons that were made there have fractional numbers.

Yes, it's quite annoying

x9136:
> Yes, it's quite annoying

If it doesn't matter to you where to create icons, then I suggest you try Figma. There is not such a wide range of work with vectors, but it may be more comfortable and faster.

PR #2027, opened 2024-04-22
Move acknowledgements entry into About screen

Goooler:
Just like what Lawnchair did.


SuperDragonXD:
I didn't notice this PR and did this myself with some UI tweaks (see the recent commits), so this PR might be somewhat redundant. Apologies for this.

PR #2093, opened 2024-05-14
+1 icon: HATSUNE MIKU, +1 link

x9136:
Fulfilling requests.

@Hamster45105 Testing #2071.

## Added
HATSUNE MIKU: COLORFUL STAGE (`com.sega.ColorfulStage.en/com.google.firebase.MessagingUnityPlayerActivity`)
## Linked
Download Navi (`com.tachibana.downloader/com.tachibana.downloader.ui.browser.BrowserIcon` → `download_navi.svg`)
## Contributor's checklist
- [x] I followed [the Lawnicons Guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md) (upd. Jan 2024) and will make changes if someone suggests. I will also make sure that Lawnicons builds correctly.

PR #2096, opened 2024-05-15
Try to fix icon checklist workflow

Hamster45105:
## Description

The icon checklist workflow [failed to run](https://github.com/LawnchairLauncher/lawnicons/actions/runs/9078779642/) on PR #2093 with error `Error: Unhandled error: HttpError: Resource not accessible by integration`, this seems to be a problem with the permissions, which is weird because it was running fine on my personal repo. Maybe there are restrictions set up on the Lawnicons repo?

I have specifically added in the permission to write a comment, so this may fix the problem. 

## Type of change

✅ Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/icon_checklist.yml
+++ b/.github/workflows/icon_checklist.yml
@@ -7,6 +7,9 @@ on:
       paths:
         - svgs/*
 
+permissions:
+  issues: write
+
 jobs:
   comment:
     runs-on: ubuntu-latest


x9136:
@SuperDragonXD Please take a look when you have time.

SuperDragonXD (approved):
GitHub actions can't really be tested unless we merge it in the repository, so Trial-and-Error will probably have to be applied here.

Also, we have branch protections applied to Lawnicons, but it shouldn't really affect PR comment writing. Not sure why it fails.

x9136:
OK, I'll make a new PR now.

x9136:
Nope. #2097

PR #2098, opened 2024-05-15
+1 icon: Quà Tặng VIP, +2 links

x9136:
Temprorary description.


SuperDragonXD:
Hmmm, seems like it didn't work either. I'll try to look into it more in the following days, since I don't have my desktop with me to troubleshoot this in-depth.

PR #2105, opened 2024-05-20
Fix tesla

peterge1998:
## Description

fix tesla.svg

Fixes #(issue)

top t line

## Type of change

### DIFF
--- a/svgs/tesla.svg
+++ b/svgs/tesla.svg
@@ -25,9 +25,9 @@
      inkscape:pageopacity="0.0"
      inkscape:pagecheckerboard="0"
      inkscape:deskcolor="#d1d1d1"
-     inkscape:zoom="2.0766337"
-     inkscape:cx="87.882613"
-     inkscape:cy="94.383519"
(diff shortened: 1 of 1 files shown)


x9136:
https://github.com/LawnchairLauncher/lawnicons/pull/1517/files#diff-4146ba9fd23928866f9190c9c1d43bcc09b7d1a2fe33e5399dbfb0df626c7a71
Tesla icon has been intentionally simplified so that its quality does not spoil on the phone screen due to too small distances between the lines at the top.

Icons with small distances between the lines look worse on the phone screen, because there they can be ~4 times smaller than here. I have redesigned hundreds of old icons to improve the overall quality of Lawnicons, although we still have quite a few icons added before May 2023, which are made without taking into account that they will be smaller on the phone screen. Sometimes deterioration in quality cannot be avoided, for example, when an icon cannot be significantly simplified without loss of recognition. But this is not the case of the Tesla icon.

In short, if you disagree, then maybe some other reviewer will accept your option.

PR #2169, opened 2024-06-19
Testing an web app icon

x9136:
Checking whether icons are applied to web applications.

### Added
ComicFury

## Contributor's checklist


x9136:
It works, but such shortcuts will always have new activity when they are added, so it's better to do without links.

PR #2176, opened 2024-06-22
Make self-review checklist collapsible

Hamster45105:
## Description

The self-review icon checklist takes up quite a bit of space on the screen, so I've made it collapsible so that it takes up considerably less space. The checklist itself is what's collapsible, not the short piece of text before it.

https://github.com/LawnchairLauncher/lawnicons/assets/133548095/6b679c87-951e-4052-ad5e-221f57ed369b

## Type of change

### DIFF
--- a/.github/icon_checklist.md
+++ b/.github/icon_checklist.md
@@ -2,6 +2,9 @@ Thanks for your contribution!
 
 While waiting for a review from our team, you can do a self-review to ensure that your icons are suitable for Lawnicons.
 
+<details>
+<summary><i>Click to expand self-review checklist</i></summary>
+
 ### Canvas and sizes
 1. Canvas: `192×192px`.
 2. Non-square icons: the long side of the icons should be `160px`.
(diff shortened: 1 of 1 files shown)


I21b (approved):

x9136:
Even when this list is open, few people use it, judging by what remains to be fixed, and if it is collapsed by default, then this is practically the same as deleting it. Therefore, this option is definitely not worth doing.

I also don't see any benefit. If you do everything as written in this list on the first attempt, then you will rarely need to scroll through it. And if you really use it and this list is collapsed, then you would have to open it more often than scroll through it. In addition, a reviewer may not be there for a week or more — while he is not there, there is rarely anything under the list for which you will need to scroll, but you may have an idea to do a self-review.

Therefore, if it were possible, then collapsing this list makes sense in such cases:
- if a collaborator has added 500 icons, then the list is not needed, because by this point he will remember it and will be able to do it properly on the first attempt;
- if a reviewer from the Lawnchair team left a review, then the list could be collapsed or even deleted automatically;
- if the list is opened by default, but it can be collapsed and this change will be saved.

Hamster45105:
Ok, fair point.

PR #2215, opened 2024-07-20
update Pi logo

KTibow:
reference logo:
image

i got to this point by messing around in inkscape (doing stuff like intersecting fill and stroke, and a little offsetting and node dragging). it's not perfect but it looks good enough, plus it's going to be displayed at a small size.

(i do realize the dimensions are a little off. i could scale it but that would make the stroke width a little off due to the way i had to make this. i know, i could've hand-made it with stroke, but this was the easiest way to make it.)


x9136:
Well, this option is certainly not suitable. Basically, the icon was made using a fill.

Screenshot 2024-07-20 at 23 17 24
The lines are too close here.

It will only do for the debug build.

KTibow:
@x9136 so add some more space then redraw with stroke?

x9136:
> so add some more space then redraw with stroke?

In addition, you need to reduce the width of the icon (it's 161.95px now) and set the correct stroke width.

KTibow:
Okay, I'll do that, but know that I'm going to keep the stroke width at 10 (so it isn't crowded) and keep the letter spacing somewhat close (so it's accurate to the original)

KTibow:
Changes made.

PR #2233, opened 2024-07-29
Clarifying the icon update template

x9136:
I deleted unnecessary info, clarified the terms and expanded the template a little. This way, we could see what needs to be changed, as well as catch activities that have been assigned to the wrong icons.

Most likely the preview is not quite accurate in the order of the blocks.
## Preview
Screenshot 2024-07-30 at 10 26 16

## Type of change

### DIFF
--- a/.github/ISSUE_TEMPLATE/icon_rebrand.yml
+++ b/.github/ISSUE_TEMPLATE/icon_rebrand.yml
@@ -20,3 +20,4 @@ body:
     id: additional-info
     attributes:
       label: Additional information
+      placeholder: Specify what else is worth considering. For example, the application has a new activity.
--- a/.github/ISSUE_TEMPLATE/icon_rebrand.yml
+++ b/.github/ISSUE_TEMPLATE/icon_rebrand.yml
@@ -20,4 +20,4 @@ body:
     id: additional-info
     attributes:
       label: Additional information
-      placeholder: Specify what else is worth considering. For example, the application has a new activity.
+      placeholder: Specify what else is worth considering. For example, the application has a new activity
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/icon_rebrand.yml
```diff
@@ -1,31 +1,24 @@
...
-
-        ```
-        App Name
-        (image to new icon)
+        Please specify what will change:
+        
+        **Old App Name**
+        The [link to the outdated icon](https://github.com/LawnchairLauncher/lawnicons/tree/develop/svgs).
+        ↓
+        **New App Name**
```
KTibow, line 15: This bit personally seems confusing. I know sometimes apps change names while rebranding, but I think it would be better to have a separate template for that.

x9136: >  I know sometimes apps change names while rebranding, but I think it would be better to have a separate template for that.

Hmm, I'll try to change it. It is rather important to find out the old name here, because many icons have several apps linked to them.

x9136: If you were referring to a separate template for the new name, then this is unnecessary. It happens too rarely and is easy to fix if there is a mistake.

SuperDragonXD (commented):
I'd suggest providing an example (either within the placeholder, or better yet, as a pre-filled value in the text box or within the instructions). The instructions doesn't really give specific formatting instructions, so this could help make requests more consistent.

SuperDragonXD (approved):

PR #2238, opened 2024-08-01
+1 icon

alufers:
# Description
Added an icon for VeloBank.

Original icon:
image

My icon:

image

The bank was renamed from Getin to VeloBank some time ago, so that's why the different package name. BTW. I am not that good at design, but I tried my best at follwing the guidelines. Any feedback is appreciated. 

### Added

VeloBank (`com.getingroup.mobilebanking`)

## Contributor's checklist


svgs/velobank.svg
x9136, line 1: Formally, this is a square icon, but it has too much emptiness, so the size should be 160×160px.

The letters should be made with a stroke.

x9136 (changes requested):

svgs/velobank.svg
alufers: Fixed

alufers:
velo_bank (1)

fixed it

PR #2239, opened 2024-08-01
+2 icons and -1 typo

5h3r10k:
# Description

Added 2 icons and fixed a typo
- Marcus
- Parkmobile

### Added

Marcus (com.marcus.android)
Parkmobile (net.sharewire.parkmobilev2)

## Contributor's checklist


5h3r10k:
Everything looks good to me now

svgs/marcus.svg
x9136, line 1: Screenshot 2024-08-03 at 13 06 17
The icon size is too small.

Screenshot 2024-08-03 at 13 05 42
The color should be black (#ffffff).

Screenshot 2024-08-03 at 13 04 13
There is a defect here.

Screenshot 2024-08-03 at 13 04 55
The corners are rounded in a different way.

It is better to make the icon with lines, rather than tracing the shape.

Considering everything:
Screenshot 2024-08-03 at 13 16 12

svgs/nj_transit.svg
x9136, line 1: The icon size is too small. It needs to be done with a stroke.

The result is something like this.
Screenshot 2024-08-03 at 13 35 43

svgs/parkmobile.svg
x9136, line 1: The icon size is too small. The color should be black (#ffffff).

svgs/snackpass.svg
x9136, line 1: The icon size is too small. The color should be black (#ffffff).

It is possible to make the icon using the base stroke width. This way it will look more appropriate.
Screenshot 2024-08-03 at 13 42 32

svgs/yi_home.svg
x9136, line 1: The icon size is too small. The color should be black (#ffffff).

x9136 (changes requested):

svgs/snackpass.svg
x9136: The previous comments remain valid. Also, all parts should be done with a stroke.

svgs/yi_home.svg
x9136: The icon size is too small. The icon needs to be done with a stroke.

x9136 (changes requested):
You have all the information to make Lawnicons-style icons: the comment from **github-actions** and the review. In this regard, it is impossible to give you more.

Good luck if you plan to do everything as required. If there is an update, I will come here in the foreseeable future, but in silence.

x9136 (changes requested):
Some of the issues have been resolved.

x9136:
The last 2 icons still don't fit.

I suggest removing them, fixing the conflicts in the branch and then everything will be fine.

5h3r10k:
I'm unsure why they don't fit, as I have kept both their sizes fitting 160x160 px.

Screenshot 2024-08-16 at 09 40 08
Screenshot 2024-08-16 at 09 42 08

Should I discard the PR and create a new one?

x9136:
> I'm unsure why they don't fit

Because icons should be made with a stroke.

5h3r10k:
I haven't ignored the message about the stroke; as is visible on the right side of the images, the fill is transparent and the stroke is black and 12px. However, I now believe that illustrator's "stroke inside" (where the stroke is entirely on the inside of the path, not on the center of the path) is causing compatibility issues. I can attempt to rework the icons in question. I will remove these 2 icons and add them in another PR.

Would you be able to show me what the icon looks like on your end? So I can troubleshoot the stroke issue.

5h3r10k:
I removed the icons in question, the rest should be good to go

x9136:
> Would you be able to show me what the icon looks like on your end? 

Screenshot 2024-08-17 at 02 31 06

5h3r10k:
My screenshots were to show you that there is no fill being applied to any of the selected paths, as is visible on the right panel. Illustrator has an option to have the stroke lie on one side of the path (not 50/50 along the path), which was used in these designs. This way, the path is actually on the edge of the shape, which makes it easy to size exactly to 160px width.

In the first screenshot, I have selected the outline guide, not the paths. Here is a better picture.

Screenshot 2024-08-16 at 19 45 56


I am unsure if the SVG is converted to a fill in other applications. I was asking if it was showing as a fill when viewed by you in your application, not from my screenshot.

x9136:
In short, the screenshot was supposed to show the following:
1. The lines where the path goes in the middle are made with the stroke.
2. The lines where there is no path in the middle are made with the fill.

5h3r10k:
So I am still waiting on my answer, does the svg show a fill when I do it this way? So I know to stick to middle stroke in future? Because when I test the APK the icon looks fine. Not opposing what you said, just trying to find a solution. Is the definition of a stroke for lawnicons where the path cannot be on the outside? If so, this should be highlighted in the guidelines.

PR #2280, opened 2024-08-23
Mark stable classes and enable Compose Compiler reports

Goooler:
https://developer.android.com/develop/ui/compose/performance/stability/fix#configuration-file


SuperDragonXD (commented):
Is there a specific advantage/reason to mark these classes as stable instead of using `kotlinx.collections.immutable.*`?

Goooler:
Compose treats immutable collections as stable, now we can mark normal collections stable instead. This will also ensure Compose's smart skipping works efficiently and reduce our package size.

SuperDragonXD (approved):
I see, thanks.

PR #2339, opened 2024-09-15
+1 icon

eifr:
# Description

add icon of one zero bank
image

### Added

App Name (`il.co.firstdigitalbank`)


x9136:
It was faster to fix than to comment. Thanks!

PR #2360, opened 2024-09-18
Update README Badges

Hamster45105:
## Description

I've updated the [shields.io](https://shields.io/) badges to look neater.

## Type of change


x9136:
In short, this update is unnecessary and not suitable.

In my understanding, neatness is expressed when you give crucial information in the most concise, effective and appropriate way. From this point of view, your option is not suitable for the following reasons:
1. I don't see any valid reasons to duplicate the repository's metrics. GitHub is already showing this information itself.
2. The new style of these badges does not suit Lawnicons, because even in the Lawnicons guidelines we have several rules for rounding icon elements.
3. You have suggested 2 new badges:
- The release badge is the third link to the releases on the same page. What's the point?
- The contributors badge is also too much. Our contributors are clearly visible on the right side of the page and they have a separate screen in Lawnicons. I didn't understand what issue you wanted to solve with this.

Hamster45105:
I have updated the PR and kept changes to a minimum, centering the badges and adding icons to all of them make it more consistent.

That being said, many projects have badges that repeat details available elsewhere, because it makes the README look more polished and provides quick at-a-glance information. Whether you agree with that or not is a matter of personal opinion.

x9136:
> That being said, many projects have badges that repeat details available elsewhere, because it makes the README look more polished. Whether you agree with that or not is a matter of personal opinion.

So we have a radically different understanding of what is neat and polished. In my understanding, polished = nothing superfluous, a deeply thought-out decision.

If you look at popular repositories, many do not overdo it with badges or with repeating the same links. Apparently both approaches are popular, if that's important to you. At the same time, the popularity of some approach does not mean that it is worth repeating.

I emphasize that it is necessary to proceed from the meaning and usefulness. What problem do you want to fix with your solution? How is your solution useful? If there was a serious answer to these questions, then we would add 20 badges and repeat the same links 10 times. For example, you recently suggested adding links to IzzyOnDroid and Obtainium. This solution is useful because it is easier to download Lawnicons through these resources than through GitHub.

README.md
```diff
@@ -1,9 +1,19 @@
 # Lawnicons
 
-[![Build debug APK](https://github.com/LawnchairLauncher/lawnicons/actions/workflows/build_debug_apk.yml/badge.svg)](https://github.com/LawnchairLauncher/lawnicons/actions/workflows/build_debug_apk.yml)
-[![Build release APK](https://github.com/LawnchairLauncher/lawnicons/actions/workflows/build_release_apk.yml/badge.svg)](https://github.com/LawnchairLauncher/lawnicons/actions/workflows/build_release_apk.yml)
```
x9136, line 4: Is there any reason to delete it?

I can imagine that developers use this badge to get to the "Build release APK" section faster.

Hamster45105: > Is there any reason to delete it?

No, just forgot to add it back in 😅

Hamster45105:
> I emphasize that it is necessary to proceed from the meaning and usefulness. What problem do you want to fix with your solution? How is your solution useful?

Yeah good points.

x9136 (approved):
Thanks!

Hamster45105:
> So we have a radically different understanding of what is neat and polished. In my understanding, polished = nothing superfluous, a deeply thought-out decision.

> I emphasize that it is necessary to proceed from the meaning and usefulness.

Yes after considering it, I agree. It does look better keeping things to a minimum.

PR #2374, opened 2024-09-24
+1 icon

Johny-adri2:
# Description
rec_room

Rec Room icon addition.

### Added

Rec Room (`com.AgainstGravity.RecRoom`)


x9136:
The stroke width and color are correctly selected, as well as the naming. The rest is not.

I recommend reading the comment above and evaluating how the icons that have already been added are made.

x9136:
In the new version of the icon, you decided to use the wrong stroke width. Also, the icon size remained unsuitable.

I fixed it for you.

PR #2423, opened 2024-10-26
Add fi.skyjake.lagrange

Sir-Photch:
# Description

### Added

App Name (`fi.skyjake.lagrange`)

Original icon for reference: https://github.com/skyjake/lagrange/blob/dev/res/lagrange-256.png


x9136:
Hey! I have taken into account the Lawnicons guidelines for you. In case of creating new icons for Lawnicons, I suggest you compare and analyze the latest changes.

PR #2434, opened 2024-11-01
Reverting the latest workflow changes

x9136:
I'm undoing the latest changes because they stop the workflow in editorial PRs, rather than skipping the app build process.

## Type of change

:white_check_mark: Bug fix (non-breaking change which fixes an issue)


x9136:
Nothing else seems to have changed, but if anything, please take a look. @Hamster45105

SuperDragonXD:
Should no longer be required after my changes in the most recent commits.

PR #2469, opened 2024-11-29
Add Cloudmoon

freakingdan:
This is my final revision for adding Cloudmoon as list of icons

Cloudmoon (1)

Edited now using Photopea

This is the best that i can do here 😓


Johny-adri2:
"I prefer my icon unique bruh", but seriously, first of all, why is it a PNG? Secondly, the outline thickness doesn't match with the guidelines. Use a VECTOR GRAPHICS program, something like Adobe Illustrator or Figma, or Inkscape… Heck, use Scratch's BUILT IN vector editor! Here's the corrected for you icon by the way.
cloudmoon
cloudmoon

x9136:
> This is the best that i can do here 😓

Even if you add the application activity to the appfilter.xml so that the icon works, it's hard for me to imagine how much the icon will differ from the rest. It strongly doesn't correspond to the Lawnicons guidelines.

> Here's the corrected for you icon by the way.

This option doesn't match the Lawnicons guidelines either.

freakingdan:
> Even if you add the application activity to the appfilter.xml so that the icon works, it's hard for me to imagine how much the icon will differ from the rest. It strongly doesn't correspond to the Lawnicons guidelines.

 :( i'm maybe not the best but i can see that this icon may not add soon enough
sorry in advance. If someone can correct my mistake here and add another PRs with the real cloudmoon icon, i will be happy ;)

freakingdan:
> Here's the corrected for you icon by the way.
It doesn't even pass the Lawnicons guidelines though ;(

PR #2470, opened 2024-11-30
+1 icon

I21b:
`P:h Diver` `com.NextFloor.phdiver`
ph_diver


x9136:
Your option is good enough to accept. Although I would look for even more opportunities to increase the space between the lines, which are now very close.

PR #2488, opened 2024-12-16
+1 icon

I21b:
`5ber.eSIM` `com.ifreegroup.moesim`

|old|new|
|---|---|


x9136:
It is better to leave only 5 and take into account the orange color in the shape of the icon to create uniqueness.

x9136:
> Like this?

Yes, but the distances between the lines should be much larger.

x9136:
x

PR #2490, opened 2024-12-20
Add 6 new icons

RaduBratan:
# Description
Added 6 new icons, mostly Romanian apps.

### Added
Ace Racer (`com.netease.racerna`)
Action Notch (`com.androxus.touchthenotch`)
Blue (`com.vms.customer`)
CFR Călători (`com.Infofer.ImtMobile`)
MyEMC (`ro.enayati.app`)
Recorder.ro(`ro.recorder.app`)


x9136:
ace_racer.svg fits, but the others don't for various reasons.

RaduBratan:
> ace_racer.svg fits, but the others don't for various reasons.

Alright, I guess I'll try to create new versions. However, would it be possible to elaborate a bit on how the icons don't fit, or give your opinion on some suggestions I made below? I feel like Recorder.ro, MyEMC and Blue at the very least accurately portray the originals while following the guidelines.

The only improvements I could think of would be:
- CFR Călători maybe doesn't need that middle line and the disconnected lower lines on the train, but the overall idea of the icon can't be improved in my opinion.
- MyEMC could also portray the name of the company (ENAYATI in caps lock underneath the shape).
- Blue appears like it's using 14px instead of 12, but I'll need to check. If it's using 12px, there's nothing I can really change, it's basically the Uber/Bolt logo but with other letters.
- The Action Notch icon could use a complete overhaul, I agree on that, it was difficult to even come up with the current design.

What do you say?

x9136:
**action_notch.svg**
> Avoid close distances between strokes.

In other words, it probably needs to be simplified somehow.
Screenshot 2024-12-31 at 18 29 52

**blue.svg**
> Avoid close distances between strokes.

The "b" is unique enough to make a high-quality icon out of it alone, rather than trying to fit 4 letters.

**cfr_calatori.svg**
> Avoid close distances between strokes.

The lines need to be moved away from each other, and, in fact, the letters are not needed.

**myemc.svg**
> Align icons to the visual center as much as possible within the guidelines.

By placing an icon of this shape in the center, it visually looks shifted downwards.

**recorderro.svg**
> No fill.

There is nothing to add.

**Why are reviews short now?**
Our bot always writes a summary of the Lawnicons guidelines. Most of reviews are explanations of which points the contributors didn't pay attention to. It takes a lot of time, and the contributors hardly study. After 100+ reviews, I can say that a different approach is better: contributors should figure it out themselves or not waste time on icons.

RaduBratan:
Thank you so much for taking your time to provide an incredibly detailed reply along with 3 possible versions of the Action Notch icon, on NYE of all days. I updated the icons to (hopefully) better fit the guidelines and will keep this PR as a draft to play a bit with some variations of CFR and MyEMC. I feel like CFR does actually benefit from having letters underneath the train (the icon looks like it's missing something without them), and I believe I still need to find the most visually appealing position for MyEMC.

Happy new year!

RaduBratan:
CFR does look better without any letters after all.

PR #2516, opened 2025-01-10
+ 1 Icon for Hive Social, the Twitter-like social media platform

ThatOneUnoriginal:
### Added
Hive Social ``(org.hiveinc.TheHive.android)``

First time contributing to this repository. I apologize if there are any blatant issues.


x9136:
The stroke width and color are chosen correctly, but there are many discrepancies with the Lawnicons guidelines.

ThatOneUnoriginal:
> The stroke width and color are chosen correctly, but there are many discrepancies with the Lawnicons guidelines.

Could you point me in the right direction for which parts are inconsistent with the lawnicon guidelines? 

This is my first time contributing to this so I may not immediately understand what is or isn't up to standards 100%

RaduBratan:
Hello. I'm not the original commenter but I could offer you some help with following the guidelines.

The first issue would be the positioning of the shapes. The content areas that you [see in the guidelines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#canvas--sizes) **have** to be touched by your shapes, meaning the mouth has to be moved lower and to the right by a few pixels and the eyes have to be moved higher and to the left by some amount too, until the edges of the lines touch the canvas limits. Considering the type of icon you created (a non-square one) and the stroke of 12px, after realignment, the width and height of the icon should be exactly 148px (160px limit - 12px stroke = 148px), whereas yours is 146px.

The second issue would be the distances between strokes. This is mentioned in passing only once [in the guidlines](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#tldr-on-icon-design) ("Avoid [...] close distances between strokes"), which is why it could be easy to miss and difficult to figure out how to implement. It could help to make the shapes slightly smaller, which combined with the aforementioned repositioning, it should increase the distances between the closest ends of any two strokes. In my opinion, a distance of 20px is the bare minimum. 

There could also be an issue with the radius of some corners, specifically two found in the small eye and one at the left end of the mouth. However, I'm not 100% sure that those corners should be rounded, because the guidelines mention that only 90deg angles should use 6-32px in radius, whereas yours aren't exactly 90deg (the eyes have obtuse angles and the mouth has an acute one). So, I don't recommend rounding anything yet.

x9136:
#2561

PR #2590, opened 2025-03-22
Avoid unnecessary svg-processor re-runs

Goooler:


Goooler:
```
diff -rq xml-before xml-after/

Files xml-before/appfilter_diff.xml and xml-after/appfilter_diff.xml differ
```

I'm unsure about `appfilter_diff.xml` changes, but it seems unrelated to this PR.

SuperDragonXD (commented):
Just to clarify, `svg-processor:run` will run when there is a change in either the `app/assets/appfilter.xml` file or the `svgs/*` folder? I can't really test this change as of now; if it does do that, you can merge this PR.

Goooler:
Correct.

PR #2617, opened 2025-04-07
Update Ente Photos

Hamster45105:
# Description

\+ 1 updated icon

### Updated

Ente Photos (`io.ente.photos.independent`)


x9136:
The icon goes beyond the content area, and it will be better if you use a 10px stroke, because this will lighten the black spots.

x9136:
Cool, thanks!

PR #2633, opened 2025-04-13
+1 icon, +1 link

glazo-0:
Added

- My NTRA (eg.gov.tra/eg.gov.tra.ui.splash.SplashActivity)


x9136:
I can repeat the same thing that I wrote last time: you have removed most of the previously added components and made an icon that almost doesn't take into account the Lawnicons guidelines.

What I recommend if you are going to add icons::
1) Analyze what changes previous PRs (with icons) have brought;
2) Fix your icon with the common issues or the Lawnicons guidelines in mind;
3) Don't create a PR until you can spot the issues as someone who has gained knowledge from 1 and 2 points.

Issue #2606, opened 2025-04-02
Known outdated icons


x9136:
### Icons

It's worth updating whenever possible.

**Ente Photos**
Now: https://github.com/LawnchairLauncher/lawnicons/blob/develop/svgs/ente_photos.svg
New: https://play-lh.googleusercontent.com/CfFyMjEYUJEnYKBy1h13-qlLa0Xn2o5Jsh7hIUNOygaH71a57Y2XLrLwQx6V4V3JrLU

**League of Comic Geeks**
Now: https://github.com/LawnchairLauncher/lawnicons/blob/develop/svgs/league_of_comic_geeks.svg
New: https://play-lh.googleusercontent.com/N2EFDeTgoOMtxe5Z1qAfClBtFS6GnckWdJxKf2bDTFtCA7utVJ1K5CZdDTCrRjEABDY

**Яндекс Go ~~ Yandex Go**
Now: https://github.com/LawnchairLauncher/lawnicons/blob/develop/svgs/yandex_go.svg
New: https://play-lh.googleusercontent.com/f_bajWOiyDWpYf-V2ric4XgP_13Tn-622VrMeXd1wBtewOc6ruqL5w7Lw1TzdFeC19E

PR #2653, opened 2025-04-23
+1 link, +3 updates

I21b:
3 of 4 in #2606
|league_of_comic_geeks|ultimate_guitar|yandex_go|
|:-:|:-:|:-:|
||||
||||
|10|12|12|

+1 link of polytone 2.0

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -132,7 +132,7 @@ dependencies {
     implementation("androidx.core:core-ktx:1.16.0")
     implementation("androidx.core:core-splashscreen:1.0.1")
     implementation("androidx.activity:activity-compose:1.10.1")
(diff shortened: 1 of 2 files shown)


x9136:
**league_of_comic_geeks**
After rounding the corners in such cases, you need to correct the size of the icon, otherwise it becomes smaller. The handle of the box is too close to the lid.

**ultimate_guitar**
The upper part has the same issue as with the handle of the previous icon. The distances between the lines should be slightly larger: 
Screenshot 2025-04-24 at 17 40 05

**yandex_go**
It is also worth working with the sizes so that the letters are closer to each other, since in the original the "G" specifically presses on the "o" due to proximity and tilt.
Screenshot 2025-04-24 at 17 48 54

PR #2657, opened 2025-04-24
Update icon_addition.md PR template

Captune:
## Description

The icon is updated, so we should use the drawable name instead of the app name.

## Type of change


x9136:
The package name is the least volatile entity. Therefore, it provides the best opportunities to search for all PRs related to a particular package. If we use the drawable name, the search results will lose relevance much faster. This will make it more difficult to track changes that may affect reviews and commits.

Captune:
> The package name is the least volatile entity. Therefore, it provides the best opportunities to search for all PRs related to a particular package. If we use the drawable name, the search results will lose relevance much faster. This will make it more difficult to track changes that may affect reviews and commits.

Thanks for your patient explanation!

PR #2665, opened 2025-05-02
+ 3 icons, + 1 update

Hamster45105:
# Description

Added 3 icons and updated 1

### Added

ABC listen (`com.thisisaim.abcradio`)
ABC NEWS (`android.AbcApplication`)
South China Morning Post (`com.scmp.newspulse`)

### Updated

ABC iview (`au.net.abc.iview`)


x9136:
In general, I agree with you.

But in particular, I don't think favicon matters if we're talking about which app it belongs to more. The ABC iview and ABC listen icons can be created based on the icon that replaces the letter "e" in their logos, but with an additional meaning (TV and headphones as an option).

x9136:
These options are good ideas, but there is an opportunity to improve the quality.

Screenshot 2025-05-04 at 14 33 27

x9136:
Thanks!

I've corrected the icons with the common issues in mind.

PR #2670, opened 2025-05-04
Create FUNDING.yml

Goooler:
Just like https://github.com/LawnchairLauncher/lawnchair/blob/15-dev/.github/FUNDING.yml

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+open_collective: lawnchair
+github: LawnchairLauncher


x9136:
I'll share my thoughts on the topic a bit.

If there is a will for Lawnicons to be well supported across many launchers and developed a bit in terms of features, there is unlikely to be any movement without sponsorship. We don't have new developers who want to do this on a regular basis. It's been that way for a long time. But we have a lot of issues of varying complexity that can be taken care of. Including those issues that were only opened in the Lawnchair repository.

Icon replenishment is in some ways secondary to whether you can install them. If an icon pack doesn't work well in a particular launcher, it doesn't matter how many icons the icon pack supports. The process of creating icons is much simpler than supporting launchers and developing features. Here sponsorship will be useful only if you want to add a lot of icons steadily and quickly. We already have enough icons to avoid chasing this in the first place. Although there are people who are willing to pay for the creation of icons, because they don't want to wait or do it themselves.

So if Lawnicons sponsorship is going to be in any way, it's primarily for developers.

PR #2900, opened 2025-07-30
initial fastlane structures from IzzyOnDroid

IzzySoft:
To prevent it from getting out of sync again…

this PR provides you with a "fastlane starter package". Some notes to sum them up in a place easy to find for you:

* the [IzzyOnDroid Fastlane Documentation](https://gitlab.com/IzzyOnDroid/repo/-/wikis/Fastlane) can guide you with maintaining the structures here
* take special care for limits, e.g. per-release changelogs can have max 500 chars, `full_description.txt` max 4,000 chars, graphics must have specific aspect ratios etc.
* you can have "bigger graphics" here; the IoD updater takes care to resize/optimize them as needed
* for the `full_description.txt` here I've used HTML compressed into a single line, to prevent the fdroid software converting each line break to a `<br>`. The tags used for this are supported by F-Droid.org as well, and to my knowledge even by PlayStore.
* whenever you add a new "type" of metadata that has not been there before (e.g. the first changelog, a featureGraphic if it wasn't yet present), please let us know so we can check and enable it.
* Details on all this in the linked documentation.
* Once merged, it will be synced at IzzyOnDroid whenever a new release is being pulled in.
* Should you have questions, be welcome to ask :wink:

And now: Enjoy!


fastlane/metadata/android/en-US/full_description.txt
```diff
@@ -1 +1 @@
-<p><i>Lawnicons</i> is a Lawnchair add-on for themed icons for selected apps, which are custom Home Screen icons that use the accent color you set in the Lawnchair settings. Originally an addon for Lawnchair 12 Alpha 5 and above to implement themed icons, it can now be used on many launchers.</p>
+<p>Lawnicons is an open-source icon pack with brand icons for Android. The icons are based on Material You, with an outlined design and rounded corners. Made by the Lawnchair team and the community.</p>
```
IzzySoft, line 1: Thanks for updating this! But wouldn't you like to keep the hint that they can be used with other launchers as well? Though I see you've put that into the shortdesc, so maybe it's not really needed here again.

x9136: Sure 👍

IzzySoft:
Ready when you are then! Note to self: as screenshots are made PNG now while keeping their names, need to cleanup the JPGs manually when pulling this in, to avoid duplication…

IzzySoft:
Thanks! Enabled on our end, screenshots cleaned up – results should become visible with the next sync around 6 pm UTC then :partying_face:

PR #2915, opened 2025-08-02
QA

x9136:
+quality
+relevance


app/assets/appfilter.xml
TechniKris, line 1: What's the reason for these removals?

x9136: In general, I delete icons when there is confirmation that apps cannot be used.

**Adobe Photoshop Mix**
https://helpx.adobe.com/mobile-apps/help/photoshop-fix-mix-eos-faq.html
> You can continue to use the versions of Photoshop Fix and Photoshop Mix already downloaded on your device until June 4, 2024. After June 4, 2024, Photoshop Fix and Photoshop Mix will no longer be operational, you'll not be able to access the apps, or your projects. Download your projects to your camera roll now to avoid losing work. 

**Tachiyomi**
https://web.archive.org/web/20250127085625/https://tachiyomi.org/news/2024-01-13-goodbye
Completely eleminated (deleted GitHub, Fdroid, etc). I only found forks with different icons.

x9136: Do I need to return something?

TechniKris: I was surprised to see this, so I wanted to make sure, so I know whether I should omit certain apps when contributing in the future. 
It seemed to me a bit unintuitive, since my assumption is that there will always be _someone_ out there who still has the app installed for one reason or another.

And a question, to have a reference point:
I know of at least a couple apps, which are either unavailable, or in a non-functional state, which still have an icon assigned to their respective components. Should those components be removed as well, in a case where their assigned icon is also used by another app?

x9136: I delete some icons not purposefully, but because I'm "on the way" when I check the relevance of our icon before working on quality. Therefore, this is the most optional task of all.

> Should those components be removed as well, in a case where their assigned icon is also used by another app?

If I came across components of dead apps, I would delete them. In some cases, the icon would have to be renamed or redesigned.

TechniKris: Got it, thanks for sharing.

PR #2934, opened 2025-08-06
SVG optimization result in KB

x9136:
Since we couldn't agree on this little thing as part of the #2932, I'm creating the PR while there's a broader discussion.

After the discussion, it will be clear what to do with the PR.


x9136:
Okay, the feature won't get a life without meaningful reasons. Useful experience for the future decisions.

If you need it, you can revive the PR. It's ready and working for today.

PR #2986, opened 2025-08-24
+2 icons, +2 links

TheWisker:
## Description

- Added new entry for *SUDOKU* pointing to existing sudoku icon.
- Added updated entry for *MGit* as its app component was outdated.
- Added two brand new icons for *UApp* and *Game Counter*.

### Added

UApp (`net.universia.ua`)
Game Counter (`net.multun.gamecounter.fdroid`)

### Linked

SUDOKU (`com.thesuncat.sudoku` → `sudoku.svg`)
MGit (`com.manichord.mg` → `com.manichord.mgit`)

#### Notes
I left the outdated entry that MGit has as I suppose removing it breaks icon support for older versions of the app that have the old app ID.

PR #2990, opened 2025-08-26
QA, -1 duplicate component

TechniKris:
## Description
  
Fixups related to #2986:
- Sorted UApp and GameCounter
- removed SUDOKU since the exact same component already exists since #2719


x9136:
You don't have to waste time on small things, because this doesn't affect the Lawnicons performanceor icon support in any way.

Tidying up can be done no more than once a year and optionally, then a significant number of duplicates may accumulate. Besides, sorting happens quite often when I add new icons.

PR #3020, opened 2025-09-08
+1 icon update

heipiao233:
## Description
Change icon of game Arknights to make it more recognizable

### Updated

Arknights (`com.hypergryph.arknights/com.u8.sdk.U8UnityContext`)


heipiao233:
It comes from the logo of the main organization in the game, Rhodes Island.
Rhodes Island

x9136:
It makes sense.

**Common issues**
2, 14, 15

Also, the icon should be made less simplified.

x9136:
arknights


I believe this option would be more recognizable. What do you think?

heipiao233:
@x9136 I think so. Sometimes in the game a single rook is also used.

PR #3037, opened 2025-09-15
updated appfilter for ParkMobile

5h3r10k:
activity name changed for parkmobile, noticed on latest update

changed to "com.parkmobile/net.easypark.android.mvp.splash.SplashActivity"


x9136:
Updating an activity breaks older apps that use the previous version. To delete activities, you need to confirm that the app is no longer working.

PR #3075, opened 2025-10-09
+1 icon

shawchip:
added MoshUp, Sajda, Anilibria App and emojimix.


x9136:
**Common issues**

anilibriaapp.svg
6
This is a regular icon, not a minimal one.
`drawable="anilibriaapp" name="AniLibria App" />` →  `drawable="anilibria" name="AniLibria" />`

emojimix.svg
7, 14

moshup.svg
7
Ideally, I would also make the height even.

PR #3084, opened 2025-10-13
Repo description update

x9136:
A description we need after the Lawnicons 2.15 release.


x9136:
> ooh, icon requests are back? how will they work?

Missing icons and components will be compiled into a request file and emailed to us.

PR #3230, opened 2025-12-14
+6 Icons, +1 Link (Fixed)

tropicalsriracha:
.


tropicalsriracha:
> BetterCounter
> Is the app version with the activity that you "fixed" no longer working?

Precisely, that's why I changed the activity. Now it works perfectly.

PR #3339, opened 2026-01-24
Use ImageVector from material icons

Goooler:


Goooler:
You are right! I found this https://www.liutikas.net/2026/02/25/org.gradle.internal.operations.trace.html

PR #3351, opened 2026-01-27
Update Open Collective funding path

Hamster45105:
## Description
  

Update Open Collective funding path


x9136:
The Lawnchair team is working on the project, so the current link cannot be replaced. It would be possible to add an additional one, but only if the experiment is successful and the following goals appear.

PR #3424, opened 2026-02-19
+1 icon

Zoomll:
This is my first icon contribution. I'm not sure if it'll be accepted immediately so I'm not going to include anymore Duolingo links for now.
This was made in Inkscape using the template by @TaivasJumala 

## Added
PAC-MAN 256 (``eu.bandainamcoent.pacman256/com.unity3d.player.UnityPlayerActivity``) > ``pac_man.svg``


x9136:
There are a couple of common issues, but more importantly, the original is turned the other way and made in a square style.

Zoomll:
I know that this updated version of the icon will still not pass due to the >8px gap, but I was unsure how I would fix it while staying accurate to the original. Do you have any ideas?

TaivasJumala:
图片

Here is 22×24 squares, to match 8px gap (considering black square as gap), each square should be 10px width(12px stroke). It will result whole icon too big.

However, what about treat I11-N12 as a stroke? Now yellow part will be the gap.

Zoomll:
Just realized I said >8px gap rather that <8px gap on accident. I think you still knew I was referring to the back part of the mouth though. 

Just to be sure, are you saying that I should close the gap there making it into a single stroke at the 1 square tall section?

Also, here's the original image I used to make the second version of the icon. It's from Wikimedia Commons. 
Original_PacMan

TaivasJumala:
yeah, close the gap, https://play.google.com/store/apps/details?id=eu.bandainamcoent.pacman256
I use the icon from playstore as reference.

Zoomll:
Here's an updated version of the icon following your suggestion.

The line is 12px thick and the rest of the icon is 8px thick.

Also, I tried to change the corner radius to 6, but it ended up looking really bad. For now I decided to just leave that part as is.

pac_man_256

Zoomll:
I realized that it could now use a 12px stoke without creating a gap below 8px. Here's that version. 

pac_man_256

I think it looks a lot better like this. I'll go ahead and commit the changes to the PR and mark it for review.

x9136:
### Common issues

**svgs/pac_man_256.svg**
2, 6, 10
For the most part, you can use [sandbox.svg](https://github.com/LawnchairLauncher/lawnicons/blob/573f61f3309f57839ff930464fbf495252789312/svgs/sandbox.svg) as a reference.

Zoomll:
I think I have corrected all the issues. Let me know if there are any others.

x9136:
QA
- No half pixels.
- No useless code (including too many points on the rounded corners).
- The inner corners are left without rounding to make the icon more expressive.

PR #3461, opened 2026-03-02
🖌️ style: Modify the HTML format to remove the blue lines between badges.

ChouChiu:
## Description

As the title 

|Before|
|---|
|image|

|After|
|---|
|image|


x9136:
Not sure if this removes the lines for all cases, but ok.

PR #3464, opened 2026-03-03
+1 Update, +1 Icon

airon-main:
### Added
Keet (`io.keet.app`)

### Updated
Super Mario Run (`super_mario_run` -> `super_mario_run`)


x9136:
### Common issues

**Alto's games**
Even if you use an alternative icon, it should still differ much more from each other, but I would prefer to try to improve the quality of the current icons.

**Rush Rally icons**
They should be different from each other. It's also better to use the "D" badge to indicate the demo version.

**super_mario_run.svg**
13
It's also worth rounding off the visor to avoid a polyline.

> [!NOTE]
> 1 icon on the first try.
> The shared queue is now at 12 icons.

airon-main:
i have a question, if i try to make the rush rally icons different with each other, the thing is, i don't know how.
if i make it simpler so that it fits the number 3 or letter O, the icon will be simpler even simpler than the icon before the update.
is making the detail small letters sufficient? (fitting the 3 or O beside the D for demo)

airon-main:
as for alto's icons, i think you can do it better than me. i'll just gonna revert it

x9136:
> i have a question, if i try to make the rush rally icons different with each other, the thing is, i don't know how. if i make it simpler so that it fits the number 3 or letter O, the icon will be simpler even simpler than the icon before the update. is making the detail small letters sufficient? (fitting the 3 or O beside the D for demo)

I need to start from scratch myself to solve this issue. Perhaps I would somehow combine the flag with "3" and "O", and add "D" separately for the demos. Or I would remove the flag by making R3 and R0. I would have to try a lot.

airon-main:
revert it is then

airon-main:
I'm pretty sure I already either reverted the icons (alto and rush rally) or fixed them (super Mario run)

x9136:
The only thing left is to fix the issue that causes the build to fail.

airon-main:
Hmm, I genuinely don't know how to fix this. Perhaps making a new branch and reapplying the changes?.

x9136:
Given the description of the error, it was probably enough to fix `rush_rally_origins_demo.svg`.

x9136:
Fixed in #3535

PR #3473, opened 2026-03-05
+3 Icons, +1 Update

airon-main:
### Added
Nino (`app.nino.mobile`)
Semeru Trans (`com.asmat.semerutrans`)
Chick-Fil-A (`chick-fil-acom.chickfila.cfaflagship`)

### Updated
Activity Launcher (`de.szalkowski.activitylauncher` -> `activity_launcher`)


x9136:
### Common issues

**svgs/activity_launcher.svg**
The meaning is eaten away. The upper part looks like a pencil, and the lower part looks like a ship's wheel.
I would make a rocket in a frame, or just a rocket if it didn't fit in a frame.

**svgs/chick_fil_a.svg**
11
Screenshot 2026-03-12 at 15 03 13
I would only use 10px for the comb.

**svgs/moememos.svg**
3, 6
It's far from the original in style, even if you make a palm instead of squares.

**svgs/nino.svg**
13, 14
It's better without rounding the corners.

**svgs/semeru_trans.svg**
It's better without rounding the corners, but I can count it as an icon made on the 1st try.

> [!NOTE]
> 1 icon on the first try.
> The shared queue is now at 13 icons.

PR #3492, opened 2026-03-13
+8 icons

StSav012:
### Added

* Dictionary (`livio.dictionary`)
* French Dictionary (`livio.pack.lang.fr_FR`)
* German Dictionary (`livio.pack.lang.de_DE`)
* Italian Dictionary (`livio.pack.lang.it_IT`)
* Portuguese Dictionary (`livio.pack.lang.pt_BR`)
* Russian Dictionary (`livio.pack.lang.ru_RU`)
* Spanish Dictionary (`livio.pack.lang.es_ES`)
* Thesaurus (`javalc6.thesaurus`)

---

The book image is taken from the [`english_dictionary`](https://github.com/LawnchairLauncher/lawnicons/blob/develop/svgs/english_dictionary.svg) drawable. Should it be improved to avoid [7 “Corner radius”](https://github.com/LawnchairLauncher/lawnicons?tab=contributing-ov-file#7-corner-radius)?

The labels are what the monochrome icons made by the app authors suggest, styled to match Lawnicons.

That's what it looks like now:
Screenshot_20260313-215228


x9136:
> Should it be improved to avoid [7 “Corner radius”](https://github.com/LawnchairLauncher/lawnicons?tab=contributing-ov-file#7-corner-radius)?

There are no issues with the `english_disctionary.svg` icon, except that I can slightly increase the consistency.

As for the icons in the PR. PRs are limited to 5 icons to prevent recurring issues and save time. I've skimmed through your icons and overall there are 3 issues: 10, 13, and the stroke weights inconsistency (which also affects the `english_disctionary.svg` icon).

english_dictionary

Updated the `english_disctionary.svg` icon; I'll add it later.

x9136:
### Common issues

**svgs/dictionary.svg**
13
Screenshot 2026-03-23 at 21 44 56

The rest of the icons need to be improved in quality, because there is no consistency between them and the letters in individual icons don't match well with each other. For example, "ES". Due to the peculiarities of the "E" and "S", the "E" seems to be larger than the "S". At the same time, the "S" is made carelessly.

StSav012:
Issue 13 for `svgs/dictionary.svg` is the main reason I've chosen a 10-px stroke for letters. I can't move the horizontal bar of the A some 5.4 px lower, for this makes it quite delta-like.

The A fitting a 8-px square inside the inner triangle.

I've made the A wider to raise the bar a bit. See the new commit.

As for the S in “ES,” that's a good point. I've missed that in actual fonts, they make the S some 2.8% taller than the E.

I'm unsure what you mean by “consistency” within a couple of letters. Of different ones. Even though it's a sans-serif font, I oppose writing the I in “IT” with a single line. Then, I can make, for instance, the T in “IT” and “PT” of the same width, but wouldn't the “IT” appear too narrow then?

Anyway, it's for you to judge. That's another try of mine.

x9136:
> I'm unsure what you mean by “consistency” within a couple of letters.

Screenshot 2026-04-01 at 19 52 54

In this situation, you need to work systematically to ensure the entire set moves in a consistent direction. You need to determine a size for all the elements that works in every case, then fine-tune the letter relationships slightly without straying too far from the overall structure. With that in mind, I established the total width for two letters, while making shapes more geometric and adjusting their relationship. At the same time, I had to slightly widen the book in one instance, which ended up making it wider for all.

If I were working on the entire set at once, then something might be different, but I'm here in the role of a reviewer, not a creator.

StSav012:
Please forgive me if I sound rude.

Letters are naturally of different widths and, as you've pointed out, heights. If you insist on making letters of the same width, then so be it.

For drawing the letters, I've taken a look at several free fonts. Below, the letters are aligned to the left.

The comparison of letters of several fonts

Even in the monospace font, the letters differ in width. In all the fonts shown above, the leg of the R spans further to the right than the apex of the oval. The Ds are wider than other letters in two of the four fonts, with the bellies stretching far to the right. Generally, all pointy elements tend to extend beyond the main area of a letter.

As for the book width, I'm on it.

StSav012:
Unless the book is square-ish enough to require the 124×142 px content area, then here it is.

x9136:
I created the icons from scratch in a general style. If any future contributions do happen, please limit each pull request to 5 icons to save time.

PR #3493, opened 2026-03-15
+7 icons, +1 link

Agroqirax:
### Added

BT Scanner (`dev.fabik.bluetoothhid`)
Idle Slayer (`com.pabloleban.IdleSlayer`)
KPN TV+ (`com.kpn.epg`)
Level Devil (`com.unept.leveldevil`)
MijnKPN (`nl.kpn.mijn`)
Next Spaceflight (`com.nextspaceflight.android.nextspaceflight`)
Polarsteps (`com.polarsteps`)

### Linked

Subway City (`com.sybogames.subway.surfers.game` → `subway_surfers.svg`)


x9136:
### Common issues

**svgs/idle_slayer.svg**
6 (a 10 px stroke weight is for special cases), 10
If it's a sword, then it's hard to recognize, you need more 45-degree angles.

**svgs/kpn_tv.svg**
6, 7, 13
It's better to leave only the crown here (if there is no separate app with this crown), rather than trying to cram everything in.

**svgs/level_devil.svg**
6, 13
It's better to leave only the little man.

**svgs/next_spaceflight.svg**
7
The fire looks ugly, including in the original. Better do something like this:
Screenshot 2026-03-22 at 19 49 58

**svgs/polarsteps.svg**
7, 13
Use 1 ellipse and 1 line for the most central part of the icon instead of the current solution.


### Links
The originals are too different from the icons you want to use. You can keep Subway Surfers City, but you need to correct the name you're suggesting.

x9136:
The `bt_scanner.svg` and `polarsteps.svg` icons are okay.

x9136:
I fixed everything for you.

Diving deeper, I found that KPN TV+ has a different icon, and your option refers to MijnKPN. Is this the case on the latest version of the app? If so, then your option should be linked to MijnKPN, and for KPN TV+ it is better to make this icon (probably):

Screenshot 2026-04-12 at 11 56 22

---

UPD: All remaining issues have been resolved. Let me know if there's anything else left.

Issue #3418, opened 2026-02-16
Release 2.18


x9136:
### Core tasks
- [ ] Add 300 new icons
- [ ] Draft the release with a description
- [ ] Test the release build

### Other issues
- [ ] Optimize Lawnicons (reduce the waiting time for icons to load). Note: my time is 5s.

### Quarterly (Jan, Apr, July, Oct)
- [ ] Add icons from the [1st try](https://github.com/LawnchairLauncher/lawnicons/issues?q=label%3A%221st%20try%22%20is%3Amerged) initiative

### Google Play description
```
- New icons, including 

Sustain Lawnicons by backing the Lawnchair team on Open Collective.
```

PR #3516, opened 2026-03-22
Remove unused strings

x9136:
We need to clean up the unused strings so that people don't translate them.

#3418


x9136:
@SuperDragonXD I started the task, but I'm not 100% aware of which strings are outdated, so it's better for you to finish it.

Q1. Is it possible to delete some folders (such as `ru`) from [the fastlane folder](https://github.com/LawnchairLauncher/lawnicons/tree/develop/fastlane/metadata/android)? If so, please delete the ones that are outdated.

Q2. Could you add a context for the lines that aren't quite obvious from the UI? For example, I don't know what `search_bar_icon_picker` is or `copied_text` is for.

PR #3542, opened 2026-04-02
+4 Icons, +1 Update

airon-main:
### Added

Centsation (`com.eipna.centsation`)
ARM-MC (`com.armmc.mobile`)
Current Activity (`io.github.ratul.topactivity`)
myITS Portal (`id.ac.its.portal`)

### Updated

PlainApp (`com.ismartcoding.plain` -> `PlainApp`)


airon-main:
image
Which one is better?

x9136:
> Which one is better?

The bottom option. For the upper one, you need to add more distance between the sectors.

x9136:
### Common issues

**arm_mc.svg**
3

**svgs/plainapp.svg**
14
The "P" needs to be shifted slightly to the right because of the void on the right.

PR #3545, opened 2026-04-02
+1 Icon

ralf401:
### Added

MOVAhome (`com.dreame.movahome`)  

Hey! Please bear with me, it is my first try :-D

I was especially unsure about the line width, 12px seems a bit bold. Looking forward what you think and what can be optimised.


x9136:
Hey!

> I was especially unsure about the line width, 12px seems a bit bold.

The fact is that 95%+ of our icons are made with a 12 px stroke weight as the base. That is, if you use something thinner, especially 8 px, then you eliminate visual consistency.

### Common issues
2, 6, 14
You also have a lot of extras in SVG, such as hidden layers and groups.

x9136:
The 14th issue has been resolved, the SVG metadata has been removed, and additional air has been pressed into the icon.

PR #3584, opened 2026-04-12
QA

x9136:
## Linked
Free Adblocker Browser (`com.hsv.freeadblockerbrowser` → `abp_for_samsung_internet.svg`)
Grindr Phone (`com.grindrapp.android` → `generic_dialer.svg`)
Rocket League Sideswipe (`com.Psyonix.RL2D` → `rocket_league_sideswipe.svg`)
Total Adblock (`com.totaladblock.adblocker` → `abp_for_samsung_internet.svg`)


svgs/osu.svg
TechniKris, line 1: ...Why? 😐

x9136: - [2 Visual balance](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#2-visual-balance)
- [4 Excessive density](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#4-excessive-density)
- [6 Text icons](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#6-text-icons)

If there is a better option, you can suggest it.

x9136: Screenshot 2026-04-12 at 16 14 07

Is this one better?

TechniKris: While I understand the desire to comply with Lawnicons' style/guidelines, in general I don't think either of these is really recognizable, especially not compared to the previous icon, which replicated the osu! cookie quite faithfully. Perhaps usage of thinner stroke weights, as per [_Stroke weights_](https://github.com/LawnchairLauncher/lawnicons/blob/develop/CONTRIBUTING.md#5-stroke-weights) could be taken advantage of here to meet Lawnicons guidelines (aside for the mild _Text icons_ guideline infraction).

For reference, here are osu!'s brand guidelines: https://osu.ppy.sh/wiki/en/Brand_identity_guidelines 
in which, as you can see, there is pretty much no other symbol or logo representing the game as a whole besides the one that is the lowercase "osu!" in a circle.

---------------

If keeping the osu! cookie as-is is not an option, then I'd like to suggest at least changing the sliders icon to have no [slider tails](https://osu.ppy.sh/wiki/en/Gameplay/Hit_object/Slider/Slidertail) (i.e. end circles) - more resembling the 2017+ default _Triangles_ skin, the 2022+ default _Argon_ skin (depicted below; in-line with the official Android client `sh.ppy.osulazer`), and the overwhelming trend of community-made skins using an invisible texture for the slider tail (plethora of examples here https://osuskins.net/ and https://skins.osuck.net/).
slider tails
However this would exclude 3 of the 4 other [game modes](https://osu.ppy.sh/wiki/en/Game_mode) from the icon. 

-----------------

In case you'd like to opt for the circle with just triangles, for the sake of keeping it game mode-agnostic, I'd suggest more overlap and size variety between the triangles, perhaps also taking advantage of the lower stroke weights. I guess the thinner stroke weight would also make the texture feel more subtle, as per `osu! Identity Guidelines.pdf`
triangles

-------------

I hope this feedback helps in any way, although with all that said I still think the plain "osu!" in a circle is the most recognizable icon, and therefore the right choice.

x9136: Thanks!

I plan to keep only 3 letters since there is no option to save the exclamation mark. Alternatively, I can keep the ellipse features, but there is a feeling that for this icon you need to keep either a full ellipse or none. If you've played osu!, would you prefer the alternative?

Screenshot 2026-04-12 at 18 04 01

TechniKris: > I plan to keep only 3 letters since there is no option to save the exclamation mark. 

Some people in the community would maul you for that 😅

But on a more serious note, 

> If you've played osu!(...)

Yes, and IMO the right one looks pretty good. It still gives the icon the distinct circular shape and feels much more recognizable. 

I really appreciate your openness to feedback and design adjustments, thanks!

PR #3602, opened 2026-04-19
feat: add Apply to Lawnchair button

wellorbetter:
## Description

Add an "Apply to Lawnchair" button to the home toolbar, allowing users to directly apply Lawnicons as their icon pack in
Lawnchair without navigating to Lawnchair settings.

This is the Lawnicons-side implementation for [LawnchairLauncher/lawnchair#6594](https://github.com/LawnchairLauncher/
lawnchair/issues/6594). The Lawnchair-side implementation is at [LawnchairLauncher/lawnchair#6643](https://github.com/
LawnchairLauncher/lawnchair/pull/6643).

- Add Palette icon for the toolbar button
- Add <queries> declaration for the intent (Android 11+)
- Show snackbar if Lawnchair is not installed

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -168,6 +168,11 @@
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
         </intent>
+        <!-- Lawnchair Apply Icons -->
+        <intent>
+            <action android:name="app.lawnchair.APPLY_ICONS" />
+            <category android:name="android.intent.category.DEFAULT" />
+        </intent>
         <intent>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.HOME" />
--- a/app/src/main/kotlin/app/lawnchair/lawnicons/ui/components/home/HomeBottomBar.kt
+++ b/app/src/main/kotlin/app/lawnchair/lawnicons/ui/components/home/HomeBottomBar.kt
@@ -33,6 +33,7 @@ import app.lawnchair.lawnicons.ui.theme.icon.Github
 import app.lawnchair.lawnicons.ui.theme.icon.IconRequest
 import app.lawnchair.lawnicons.ui.theme.icon.LawnIcons
 import app.lawnchair.lawnicons.ui.theme.icon.OpenCollective
+import app.lawnchair.lawnicons.ui.theme.icon.Palette
 import app.lawnchair.lawnicons.ui.theme.icon.Search
 import app.lawnchair.lawnicons.ui.util.Constants
 import app.lawnchair.lawnicons.ui.util.PreviewLawnicons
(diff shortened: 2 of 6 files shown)


x9136:
Since you're forcing this issue, I need to redesign the toolbar, because 6 elements is too much.

There are questions. 

1) How will it work for other icon packs? It's unlikely that you will add buttons to each one.
2) Will this only work for Lawnchair or for any launcher that allows you to apply icons from the icon pack app?

wellorbetter:
> Since you're forcing this issue, I need to redesign the toolbar, because 6 elements is too much.
> 
> There are questions.
> 
> 1. How will it work for other icon packs? It's unlikely that you will add buttons to each one.
> 2. Will this only work for Lawnchair or for any launcher that allows you to apply icons from the icon pack app?

I agree the toolbar might be getting crowded. I haven't contributed to Lawnicons before — this PR was mainly to validate the
Lawnchair-side implementation. Do you have any suggestions on where to place it or how to approach this?

For context, the flow is: user taps the button → Lawnchair receives the intent and shows a confirmation BottomSheet on its side → user confirms or dismisses.

x9136:
>  Do you have any suggestions on where to place it or how to approach this?

Yes, I have an idea, but if the whole feature is applicable not only for Lawnchair, then this should be taken into account. For example:
- If I can apply icons from the Lawnicons app in the Smart Launcher, then the text linked to the Lawnchair is no longer suitable. 
- If I can apply icons from the Lawnicons app only in Lawnchair, then it's better to hide this action in Lawnicons if Lawnchair isn't installed, rather than showing the snackbar.

Also, speaking of other icon packs, if the refinement in Lawnchair leads to the pop-up being shown in other icon packs, then it's better to do the feature itself differently, so the refinement of Lawnicons may become different.

wellorbetter:
> > Do you have any suggestions on where to place it or how to approach this?
> 
> Yes, I have an idea, but if the whole feature is applicable not only for Lawnchair, then this should be taken into account. For example:
> 
> * If I can apply icons from the Lawnicons app in the Smart Launcher, then the text linked to the Lawnchair is no longer suitable.
> * If I can apply icons from the Lawnicons app only in Lawnchair, then it's better to hide this action in Lawnicons if Lawnchair isn't installed, rather than showing the snackbar.
> 
> Also, speaking of other icon packs, if the refinement in Lawnchair leads to the pop-up being shown in other icon packs, then it's better to do the feature itself differently, so the refinement of Lawnicons may become different.

Makes sense — a generic "Apply" entry that detects which launchers support it at runtime, shows a picker if multiple are available, and hides itself when none are installed.

This would mean maintaining a list of supported launcher intents on the Lawnicons side, and each launcher needs to handle the intent on their end. Currently only Lawnchair supports this, but the approach would be extensible.

Also, where would you suggest placing the entry?

x9136:
> Also, where would you suggest placing the entry?

I was planning this revision to clean up the Lawnicons UI. It's not entirely about your improvement, but part of it is.

PR #3603, opened 2026-04-19
+1 icon, +1 link

Ashton20271:
map the new amethyst launcher id to the existing icon


x9136:
Just to confirm my understanding: one of the app IDs corresponds to the Amethyst release build, and the other to the debug build. Is that correct?

If so, please specify which app ID applies to the debug build, because I have conflicting info.

Ashton20271:
the icon that already exists is mapped to the debug build (org.angelauramc.amethyst.debug/net.kdt.pojavlaunch.TestStorageActivity) the non debug one is org.angelauramc.amethyst/net.kdt.pojavlaunch.TestStorageActivity which removes the .debug as you can see in https://github.com/AngelAuraMC/Amethyst-Android/releases there are 2 options, one for the debug version and one for the stable version

x9136:
Icons of different builds should have different icons, so I added a separate icon for the Amethyst debug build.

PR #3614, opened 2026-04-21
+1 icon

mattague:
### Added
Absorb (`com.barnabas.absorb` → `generic_voice.svg`)


x9136:
A unique icon is needed here.

mattague:
The icon is almost identical, to the point that I think a custom icon would be more cluttered. The original icon is the same general design, with slightly more vertical bars, which fade to match the background on the ends. If this still requires a custom icon, I will work on that tonight. 

[icon.png](https://github.com/user-attachments/assets/e2a7c009-fc64-4633-8585-844c9a5e4388)

x9136:
Your comment coincides with my review assessment, but not with the conclusions. The original icon is unique enough to be created.

x9136:
I made the icon for you because I had a rare chance to create a moment where there are no open PRs with icons.

PR #3672, opened 2026-04-30
Change debug build skip logic for meta changes

### LINKED ISSUE #2264: Stop the debug apk build when there are no changes
### About this discussion

It takes several minutes to build the debug apk, but if Lawnicons doesn't change, then the build action is triggered in vain.

I suggest adding a check to the build action so that it is successfully canceled when there are no updates. This would help in some cases to merge the PR without waiting for the build.

x9136:
- Use include-based path filtering to skip debug builds on meta changes
- Restore notifications so they always send, regardless of build status

For meta-changes, the checks will be 4 times faster.

Closes #2264

### DIFF
--- a/.github/workflows/build_debug_apk.yml
+++ b/.github/workflows/build_debug_apk.yml
@@ -17,21 +17,22 @@ jobs:
         with:
           submodules: true
           fetch-depth: 0
-      - name: Check for ignored paths
+      - name: Check for changes in source files
         uses: dorny/paths-filter@v4
         id: filter
         with:
-          base: ${{ github.event_name == 'push' && github.event.before || github.event.pull_request.base.sha }}
           filters: |
             workflows:
-              - '!**.md'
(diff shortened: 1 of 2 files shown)


Goooler:
Why dno't move the job out, you can reuse it by `workflow_call`, there is an example: https://github.com/Goooler/homebrew-repo/blob/59242f2ddc9e478cb2a6e9b74b1e844f375bed8b/.github/workflows/check-packages.yml#L10-L13

x9136:
> Why dno't move the job out, you can reuse it by `workflow_call`

I don't see where it's being duplicated now or where it needs to be reused in the near future (if you're talking about `check-paths`).

What's much more important after nearly 2 years is finally not running debug builds on every single change.

SuperDragonXD (approved):
This should be fine for now.

If this still doesn't work, I guess we can just make another PR until it does work.

PR #3835, opened 2026-06-10
+1 icon, +1 update

mcbabo:
added meine_sv
updated refra (former gallery)

### Added
MeineSV (`at.itsv.mobile.meinesvstatus` → `meine_sv.svg`)
3 × ReFra (`com.dot.gallery` → `refra.svg`)


x9136:
svgs/meine_sv.svg
It is better to make 1 line in the center, rather than separating the shapes from each other. Plus, it is necessary to avoid broken curves.

svgs/refra.svg
The corners should be sharper.

PR #3974, opened 2026-08-02
Fix clock icons not showing on Moto Launcher

diredocks:
## Problem

On Motorola devices, the stock launcher (`com.motorola.launcher3`) treats clock apps specially: it wraps their icon in a ClockDrawableWrapper that casts the adaptive icon's foreground to a LayerDrawable.

Lawnicons generated clock icons with a plain <inset> foreground, which inflates to an InsetDrawable. The cast failed with a ClassCastException, which made the launcher fall back to the stock system icon — so Lawnicons clock icons were never displayed.

## Fix

In SvgFilesProcessor.createAdaptive, wrap the `<inset>` foreground/monochrome in a `<layer-list><item>`. The foreground now inflates to a LayerDrawable, matching what the launcher expects.

```xml
<!-- before -->
<foreground><inset android:inset="32%" .../></foreground>

<!-- after -->
<foreground>
  <layer-list><item><inset android:inset="32%" .../></item></layer-list>
</foreground>
```

This applies to all icons uniformly, and the visuals are unchanged since the layer-list contains only one item with the same inset.

## Verification

The root-cause analysis was done with LLM assistance by inspecting decompiled launcher sources and logcat. The fix itself was built, installed on a Motorola device, and confirmed working there: the Lawnicons clock icon now displays and no ClassCastException appears in logcat.

photo_2026-08-02_18-00-12

### DIFF
--- a/svg-processor/src/main/kotlin/app/lawnchair/lawnicons/helper/SvgFilesProcessor.kt
+++ b/svg-processor/src/main/kotlin/app/lawnchair/lawnicons/helper/SvgFilesProcessor.kt
@@ -117,13 +117,15 @@ object SvgFilesProcessor {
             .addAttribute("xmlns:android", "http://schemas.android.com/apk/res/android")
         root.addElement("background")
             .addAttribute("android:drawable", bgColor)
-        root.addElement("foreground").addElement("inset")
+        root.addElement("foreground").addElement("layer-list").addElement("item")
+            .addElement("inset")
             .addAttribute("android:inset", "32%")
             .addAttribute(
                 "android:drawable",
                 "@drawable/" + FilenameUtils.getBaseName(foregroundXml),
             )
-        root.addElement("monochrome").addElement("inset")
+        root.addElement("monochrome").addElement("layer-list").addElement("item")
+            .addElement("inset")
             .addAttribute("android:inset", "28%")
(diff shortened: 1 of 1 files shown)


x9136:
Thanks!

Do you think this will work if we add dynamic clock support in the future?
https://kvaesitso.mm20.de/docs/developer-guide/integrations/icon-packs.html#clock-icons

@SuperDragonXD The fix was confirmed by people on Discord.

diredocks:
I don't know much about how dynamic clock icons work on Android – seems launcher-specific. I briefly looked into it while working on this PR, but Lawnicons doesn't support dynamic clocks yet, so I didn't dig deeper.

That said, this PR should be fine if you add dynamic clock support later – it's just XML restructuring and shouldn't break anything.

When you do get to implementing it, feel free to ping me and I'll help test.

SuperDragonXD (approved):

Donnnno:
This is what I've been searching for for a very long time now! Great that there's a fix for this issue.
Out of curiosity, is this tested with different launchers to see if it works everywhere?

x9136:
> Out of curiosity, is this tested with different launchers to see if it works everywhere?

I haven't tested it, but Lawnicons Nightly with  the fix is available.

